PR #4, opened 2023-02-24
fix: Don't use hyphen in package name

lynnagara:


untitaker (approved):
unsure why this is a problem. the python sdk also has a hyphen

lynnagara:
I think this isn't the problem I thought it was, closing

PR #5, opened 2023-02-24
build: Include yaml files in Python package

lynnagara:


evanh (approved):

untitaker (approved):

untitaker:
```
$ python setup.py sdist
error: Multiple top-level packages discovered in a flat-layout: ['topics', 'schemas', 'sentry_kafka_schemas'].

To avoid accidental inclusion of unwanted files or directories,
setuptools will not proceed with this build.

If you are trying to create a single distribution with multiple packages
on purpose, you should not rely on automatic discovery.
Instead, consider the following options:

1. set up custom discovery (`find` directive with `include` or `exclude`)
2. use a `src-layout`
3. explicitly set `py_modules` or `packages` with a list of names

To find more information, look for "package discovery" on setuptools docs.
```

untitaker:
ah that's resolved with https://github.com/getsentry/sentry-kafka-schemas/pull/6

PR #14, opened 2023-03-01
test: Add test that loads all topics

lynnagara:


python/tests/test_valid_topic_data.py
```diff
@@ -0,0 +1,13 @@
+from pathlib import Path
+
+from sentry_kafka_schemas import get_schema
+
+
+def test_all_topics() -> None:
+    topics_dir = Path.joinpath(Path(__file__).parents[2], "topics")
+    for filename in topics_dir.iterdir():
+        if filename.suffix != ".yaml":
```
evanh, line 9: No support for `.yml` 😆 ?

evanh (approved):

untitaker (approved):

python/tests/test_valid_topic_data.py
lynnagara: Trying to keep things opinionated over here...

PR #23, opened 2023-03-02
feat: Add topic description

lynnagara:
@untitaker Just an idea, how about we add a description field to the topic so people can understand at a glance what it's for.

The topic names, especially around metrics, can be extremely unintuitive. Since these can be hard to actually change (many services and deployments involved), let's try and augment this with a more easy to understand description?


untitaker (approved):
Sure why not

PR #2, opened 2022-12-21
feat: Add topic and schema definition

lynnagara:
This PR adds the basic structure for the schemas repository outlined in https://github.com/getsentry/rfcs/blob/main/text/0072-kafka-schema-registry.md. Each topic is a yaml file in the `topics` directory. Each schema is a file in the `schemas` directory. Example payloads can be added as files in the `examples` directory. Topics can reference schemas and examples.

The example provided is for the `querylog` topic which is used by Snuba.

PR #24, opened 2023-03-02
ref: Add Python types

untitaker:
* I figured out that repeated invocations of `make install` were failing
  due to a random bug in setuptools. Add a pyproject.toml to rectify
  this.
* Add jsonschema-gentypes in a new requirements.txt. Add a new script
  that calls it for all schemas.
* I couldn't figure out a way to run code generation as part of
  `setup.py sdist` or `setup.py bdist_wheel`. The commands can be
  overridden from setup.py, but there is no way to install
  jsonschema-gentypes before those run.

  So now our build process runs entirely through the makefile which
  explicitly installs those dependencies first.

* I tried using the resulting schema files in Snuba, the main ergonomic
  problem is that the generated schemas do not distinguish between
  required and optional attributes. I think one could fix that.

* The other problem is that there's no nameable type for parts of the
  schema. The errors processor wants a nameable type for parts of the
  event, and the generated names are neither pretty nor stable.

## Why types?

our goal for all of these schemas is to:
* not reject any existing messages in prod, because we (sort of) know those are valid (= the schema should not be too strict)
* reject all messages that would later make the consumer crash (=the schema should not be too permissive)

right now, we have a process for `#1`: we run it in prod and if the validation errors stop, we are done

we don't have a process for `#2`.

i.e., one can write an overly permissive wildcard schema today, and we will ship it, but we have no way of knowing whether malformed messages will actually be caught by it

afaict the only way to do that statically, is to have the entire consumer be properly typed and then compare its python types against what the schema produces (see: jsonschema-typed package)

now obviously we are far away from typing our codebase outside of snuba. so how can we ensure our schemas are not overly permissive? can we do it at runtime somehow?

ideas discussed in slack were around fuzzing consumers using something like hypothesis-jsonschema, but we figured that exporting python types is simpler


lynnagara (approved):
Hell yeah 💯

PR #26, opened 2023-03-03
feat: Implement `get_schema` in Rust

lynnagara:
Basic rust implementation. Doesn't handle schema versions or parsing json schemas yet.

### DIFF
--- a/rust/lib.rs
+++ b/rust/lib.rs
@@ -70,6 +70,9 @@ pub struct Schema {
     schema: String,
 }
 
+/// # Errors
+///
+/// Will return `Err` if `topic` or `version` is not found or if schema data is invalid.
 pub fn get_schema(topic: &str, _version: Option<u16>) -> Result<Schema, SchemaError> {
     let topic_path = format!("./topics/{}.yaml", topic);
     let mut topic_data = TopicData::load(&topic_path)?;
--- a/rust/lib.rs
+++ b/rust/lib.rs
@@ -2,14 +2,14 @@ use serde::{Deserialize, Serialize};
 use std::fmt;
 use std::fs::{read_to_string, File};
 
-#[derive(Clone, Debug, PartialEq, Serialize, Deserialize)]
+#[derive(Debug, PartialEq, Serialize, Deserialize)]
 #[non_exhaustive]
 enum SchemaType {
     #[serde(rename = "json")]
     Json,
 }
 
-#[derive(Clone, Debug, PartialEq, Serialize, Deserialize)]
+#[derive(Debug, PartialEq, Serialize, Deserialize)]
 #[non_exhaustive]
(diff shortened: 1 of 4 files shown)


rust/lib.rs
```diff
@@ -1,8 +1,92 @@
...
+        return Err(SchemaError::InvalidType);
+    }
+
+    let json_schema_path = format!("./schemas/{}", latest.resource);
+
+    Ok({
+        Schema {
+            version: latest.version,
+            schema_type: latest.schema_type.clone(),
+            compatibility_mode: latest.compatibility_mode.clone(),
```
lynnagara, line 86: Not sure if there's an easy way to avoid these clones

rust/lib.rs
```diff
@@ -1,8 +1,92 @@
...
+    compatibility_mode: CompatibilityMode,
+    // TODO: Actually parse the json schema
+    schema: String,
+}
+
+pub fn get_schema(topic: &str, _version: Option<u16>) -> Result<Schema, SchemaError> {
+    let topic_path = format!("./topics/{}.yaml", topic);
+    let topic_data = TopicData::load(&topic_path)?;
+    // TODO: Respect version
+    let latest = topic_data.schemas.last().unwrap();
```
untitaker, line 75: i think this code should be panic-free, we can't afford to panic in library code that is depended on by random services imo. you can do `topic_data.schemas.last().ok_or(SchemaError::InvalidType)?`

rust/lib.rs
untitaker: add `derive(Clone)` to the CompatibilityMode enum, but i think you should be able to remove the `.clone()` already anyway? It will mean you can't use `latest` afterwards but you're returning anyway

rust/lib.rs
```diff
@@ -1,8 +1,92 @@
...
+        serde_yaml::from_reader(f).map_err(|_| SchemaError::TopicNotFound)
+    }
+}
+
+#[derive(Debug, PartialEq, Serialize, Deserialize)]
+pub struct Schema {
+    version: u16,
+    schema_type: SchemaType,
+    compatibility_mode: CompatibilityMode,
+    // TODO: Actually parse the json schema
```
untitaker, line 67: this might be fine if we're not in the business of prescribing a specific schema validation library. otherwise I would suggest to make schema an enum like:

```
#[serde(tag = "schema_type")]
pub enum Schema {
    Json {
        schema: ...
    },
    DingusRPC {
        schema_with_some_other_type: ...
    }
}
```

untitaker (approved):

rust/lib.rs
untitaker: ah, you need to replace `last()` with something like `pop()` to make `latest` of type `Schema` instead of `&Schema`

PR #31, opened 2023-03-07
fix: Fix event schema wrt nullable tag keys

untitaker:


lynnagara:
do you know who's producing this garbage and why? 🙊

lynnagara (approved):

untitaker:
> do you know who's producing this garbage and why? aren't tags only allowed to be strings in sentry? 🙊

It's relay ;/

PR #33, opened 2023-03-08
feat: Add basic iter_examples function

untitaker:
Also merge the two testsuites because the code duplication was getting
ridiculous.


python/sentry_kafka_schemas/sentry_kafka_schemas.py
```diff
@@ -121,3 +124,26 @@ def get_schema(topic: str, version: Optional[int] = None) -> Schema:
     cached_schema = __TOPIC_TO_SCHEMA[schema_key]
     assert cached_schema is not None
     return cached_schema
+
+
+def iter_examples(topic: str, version: Optional[int] = None) -> Iterable[Path]:
```
lynnagara, line 129: Could we return the actual examples instead of the paths? Seems like one less thing the end user would need to do

untitaker: I wanted to have useful output when running `pytest tests/test_examples.py`. I am now returning a more complex object, ptal

lynnagara: A bit of a nitpick but I'm not sure I see the value in returning the `schema` inside each example. Can't the tests easily get hold of that get that via `get_schema(topic, version)`

untitaker: fair enough, done!

lynnagara (approved):

PR #41, opened 2023-03-14
feat(Querylog): Organization ID Field

rahul-kumar-saini:
Context in PR description [here](https://github.com/getsentry/snuba/pull/3857).


lynnagara (approved):

lynnagara (changes requested):
Actually looking at the reference PR again, should it be integer or null?

untitaker (commented):
would prefer if you added test payloads that reflect all possible states

lynnagara (approved):

PR #49, opened 2023-03-20
feat(metric alerts): Add schema, topic, and example

ceorourke:
Add a schema, topic, and example for the query subscription consumer. See the schema [here](https://github.com/getsentry/sentry/blob/master/src/sentry/snuba/json_schemas.py) for reference.


topics/query-subscription-consumer.yaml
```diff
@@ -0,0 +1,9 @@
+topic: query-subscription-consumer
+description: Metric alert consumer
+schemas:
+  - version: 1
+    compatibility_mode: none
+    type: json
+    resource: query-subscription-consumer.schema.json
+    examples:
+      - query-subscription-consumer/1/
```
lynnagara, line 9: The query subscription consumer is just one client of this schema, I think it's better to avoid this consumer's name should be written anywhere in the schema.

examples/query-subscription-consumer/1/query-subscription-consumer
```diff
@@ -0,0 +1,19 @@
+{
+  "version": 3,
+  "payload": {
+    "subscription_id": "1234", 
+    "result": {
+      "data": [
+        {
+          "hello": 50,
```
lynnagara, line 8: Is this real data? This does not seem like anything that should be published on this topic. Isn't there supposed to be a "count" here?

examples/query-subscription-consumer/1/query-subscription-consumer
```diff
@@ -0,0 +1,19 @@
+{
+  "version": 3,
+  "payload": {
+    "subscription_id": "1234",
```
lynnagara, line 4: Is this the actual format of a subscription ID? Isn't it supposed to look like a partition number and UUID

lynnagara:
What was the process you used to create this schema?

topics/query-subscription-consumer.yaml
```diff
@@ -0,0 +1,9 @@
+topic: query-subscription-consumer
```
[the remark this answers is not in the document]
ceorourke: Should I make separate files for each topic, or can they all be in the same file?

ceorourke:
> What was the process you used to create this schema?

I wrote it manually, referencing the existing schemas in this repo and the schemas [here](https://github.com/getsentry/sentry/blob/master/src/sentry/snuba/json_schemas.py).

untitaker:
>I wrote it manually, referencing the existing schemas in this repo and the schemas [here](https://github.com/getsentry/sentry/blob/master/src/sentry/snuba/json_schemas.py?rgh-link-date=2023-03-21T19%3A55%3A55Z).

I think lyn meant, what did you look at to determine which fields should exist in the schema. Did you look at the code that consumes those messages, vs tests that contain example payloads, vs something else

topics/query-subscription-consumer.yaml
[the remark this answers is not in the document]
lynnagara: Each topic should be it's own file, they can share the schema though

topics/query-subscription-consumer.yaml
lynnagara: Are you able to provide one example for each topic? ...or is that too much?

topics/query-subscription-consumer.yaml
```diff
@@ -0,0 +1,9 @@
+topic: events-subscription-results
+description: Metric alert events consumer
+schemas:
+  - version: 1
+    compatibility_mode: none
+    type: json
+    resource: query-subscription-consumer.schema.json
```
lynnagara, line 7: Also shouldn't be named after one of the clients. Recall that the producer will use this too when it is publishing the messages. You can call it `subscription-results.schema.json`

schemas/query-subscription-consumer.schema.json
```diff
@@ -0,0 +1,77 @@
+{
+  "$schema": "http://json-schema.org/draft-07/schema#",
+  "$ref": "#/definitions/Main",
+  "definitions": {
+    "Main": {
+      "title": "Subscription Wrapper Schema",
```
lynnagara, line 6: Please see https://github.com/getsentry/sentry-kafka-schemas#using-python-types for how the `title` property should be used

ceorourke: I don't think I quite understand what I need to do from reading that documentation. Does every property need a title? Also I was a bit confused by the path listed, imo it should say `sentry-kafka-schemas/python/sentry_kafka_schemas/schema_types/`.

lynnagara: Just the top level one.. and optionally anything else that you wish to generate a Python TypedDict for.

schemas/subscription-results.schema.json
```diff
@@ -0,0 +1,79 @@
+{
+  "$schema": "http://json-schema.org/draft-07/schema#",
+  "$ref": "#/definitions/Main",
+  "definitions": {
+    "Main": {
+      "title": "subscription-results",
```
untitaker, line 6:
```suggestion
      "title": "subscription_results",
```

examples/query-subscription-consumer/1/subscription-results
```diff
@@ -0,0 +1,35 @@
...
+      }, 
+      "trace_output": ""
+    },
+  "request": {
+    "query": "MATCH (events) SELECT count() AS `count` WHERE type = \'error\' AND project_id IN array(4551586472787969)", 
+    "tenant_ids": {
+      "organization_id": "<subscriptions>", 
+      "referrer": "subscriptions_executor"
+    }
+  }
```
untitaker, line 31: you've got a comma missing here

untitaker (commented):
otherwise lgtm modulo test failures

ceorourke:
Minor feedback - this f string isn't working, it just prints out `definition_name` instead of the variable: https://github.com/getsentry/sentry-kafka-schemas/blob/main/python/tests/test_valid_schemas.py#L57

schemas/subscription-results.v1.schema.json
```diff
@@ -0,0 +1,125 @@
+{
+   "$schema":"http://json-schema.org/draft-07/schema#",
+   "$ref":"#/definitions/Main",
+   "definitions":{
+      "Main":{
+         "title":"subscription_results",
```
lynnagara, line 6: @untitaker This type doesn't seem to be getting generated correctly

untitaker: what is wrong about it? I'm looking at `SubscriptionResults` in the resulting schema, and it looks like this:

```

class SubscriptionResults(TypedDict, total=False):
    """subscription_results."""

    version: Required[int]
    """required"""

    payload: Required[Union["PayloadV3", "PayloadV2"]]
    """required"""


```

topics/generic-metrics-sets-subscription-results.yaml
```diff
@@ -0,0 +1,9 @@
+topic: generic-metrics-sets-subscription-results
```
lynnagara, line 1: This sets/distributions/counters split of generic metrics doesn't seem right to me. I'd like to get that changed. Wanna just omit this from here for now? We'll try and fix that then add it later?

ceorourke: Sure I can take them out 👍

ceorourke:
These tests are passing for me locally, but I had to re-run `make build`. I'm testing using it locally in this PR: https://github.com/getsentry/sentry/pull/46296

untitaker:
sorry about that, my "bespoke linting test" crashed on that schema, pushed a fix

lynnagara:
How come you changed the timestamp in the example from integer to string? Where did you pull that example from? Are you sure it's correct?

ceorourke:
> How come you changed the timestamp in the example from integer to string? Where did you pull that example from? Are you sure it's correct?

It was a mistake before, we're actually sending a string along. The example is from me running it locally, and it lines up with existing tests we have too.

schemas/subscription-results.v1.schema.json
```diff
@@ -0,0 +1,70 @@
...
+                              }
+                           }
+                        }
+                     },
+                     "required":[
+                        "data"
+                     ]
+                  },
+                  "timestamp":{
+                     "type":"string"
```
ceorourke, line 48: It looks like the specific format of datetime string we send along doesn't match any of the defined formats: https://json-schema.org/understanding-json-schema/reference/string.html#dates-and-times

untitaker: I believe this is `"date-time"`, or at least I am pretty sure what you're sending is RFC 3339. I don't think the examples on that doc page are exhaustive.

ceorourke: I had that before and was seeing validation errors :/

lynnagara: I think we are sending iso 8601. 3339 is stricter and requires the offset

untitaker: the set of allowed formats in ISO 8601 is vast. Can we change the code to just send RFC 3339? I believe it is a strict subset. But I am fine with `string` as well

ceorourke: 🤷‍♀️ string is okay with me if it's okay with you too

lynnagara: I wouldn't mind taking the opportunity to standardise the way we do datetime... The question is: if snuba starts sending the new format will Sentry's results consumer be able to handle it right away, or is this going to be a painful synchronization process?

untitaker: let's merge string, idk if it's worth fixing rn...

untitaker:
brace yourself for the autoformatting bot

schemas/subscription-results.v1.schema.json
lynnagara: sgtm

lynnagara (approved):

PR #51, opened 2023-03-22
ref: Add autoformatting for JSON

untitaker:


untitaker:
as noted on https://github.com/getsentry/action-github-commit -- need to use a different GITHUB_TOKEN. probably using https://github.com/getsentry/action-github-app-token

.github/workflows/ci.yml
```diff
@@ -7,6 +7,33 @@ on:
...
+          private_key: ${{ secrets.SENTRY_INTERNAL_APP_PRIVATE_KEY }}
+
+      - uses: actions/setup-python@v2
+        with:
+          python-version: 3.8
+
+      - run: make install format
+
+      - uses: getsentry/action-github-commit@bec0adb2924f8c19a9f84c18c83d375caf02ca38 # main
+        if: github.ref != 'refs/heads/master'
```
lynnagara, line 32: does it need to be main?

untitaker: you are right! fixed

Makefile
```diff
@@ -23,6 +24,8 @@ install: python/sentry_kafka_schemas/schema_types
 
 format:
 	black python/
+	cargo fmt
+	yarn prettier --write .
```
lynnagara, line 28: Is it possible to do something like:
```suggestion
	yarn prettier --write **.*.json
```

lynnagara (approved):

Makefile
untitaker: @lynnagara yeah I would say whatever prettier supports is fair game. It doesn't seem to touch files it doesn't understand. Unlikely it conflicts with black :D

PR #55, opened 2023-03-24
ref(outcomes): Add outcomes schema

john-z-yang:
### Overview
Add schema for outcome for validation


untitaker (commented):
please add 1-2 example payloads from local development, also a yaml file in `topics/` to describe the topic

schemas/outcomes.v1.schema.json
```diff
@@ -0,0 +1,45 @@
...
+  "required": [
+    "timestamp",
+    "org_id",
+    "project_id",
+    "key_id",
+    "outcome",
+    "reason",
+    "event_id",
+    "category",
+    "quantity"
```
lynnagara, line 43: Are they all required?

lynnagara (approved):

schemas/outcomes.v1.schema.json
john-z-yang: I'm not sure, what can I do to verify what fields are required and which are not?

PR #56, opened 2023-03-26
feat: Add ingest-metrics topic

untitaker:
ingest-metrics is technically two topics, but Relay treats it as one. The messages are shaped exactly the same.

### DIFF
--- a/schemas/ingest-metrics.v1.schema.json
+++ b/schemas/ingest-metrics.v1.schema.json
@@ -65,11 +65,8 @@
     },
     "StringToString": {
       "type": "object",
-      "additionalProperties": false,
-      "patternProperties": {
-        "^.*$": {
+      "additionalProperties": {
           "type": "string"
-        }
       }
     }
   }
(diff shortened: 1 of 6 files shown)


lynnagara:
What does "technically two topics" exactly? Does some other `ingest-generic-metrics` topic also exist?

schemas/ingest-metrics.v1.schema.json
```diff
@@ -0,0 +1,73 @@
...
+        "value": {
+          "anyOf": [
+            {
+              "type": "integer"
+            },
+            {
+              "type": "array",
+              "items": {
+                "type": "number"
+              }
```
jjbayer, line 43: Is there a way to tell json schema that we have either `"type": "c"` with `"value": <int>` or `"type": "^[ds]$"` with `"value": <array>`? Maybe by having two different subschemas?

untitaker:
@lynnagara yes, ingest-generic-metrics exists for performance but it's IMO the same "logical" topic with the same exact schema.

schemas/ingest-metrics.v1.schema.json
untitaker: I believe that might be possible using `allOf`, but I am not sure we want to use that construct (because it produces worse types). Also technically the metrics indexer does not care about it. We could consider it for the snuba topics

jjbayer (approved):

lynnagara (approved):

PR #65, opened 2023-03-28
fix: Subscription result schema

lynnagara:
More typing for the result, so this can be used in the Snuba producer

### DIFF
--- a/schemas/subscription-results.v1.schema.json
+++ b/schemas/subscription-results.v1.schema.json
@@ -3,7 +3,7 @@
   "$ref": "#/definitions/Main",
   "definitions": {
     "Main": {
-      "title": "subscription_results",
+      "title": "subscription_result",
       "type": "object",
       "additionalProperties": false,
       "properties": {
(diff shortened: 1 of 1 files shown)


schemas/subscription-results.v1.schema.json
```diff
@@ -3,7 +3,7 @@
   "$ref": "#/definitions/Main",
   "definitions": {
     "Main": {
-      "title": "subscription_results",
+      "title": "subscription_result",
```
lynnagara, line 6: Renamed to singular for consistency since this represents a singular result

schemas/subscription-results.v1.schema.json
```diff
@@ -27,17 +27,37 @@
...
+                  "type": "object"
+                },
+                "profile": {
+                  "type": ["object", "null"]
+                },
+                "trace_output": {
+                  "type": "string"
                 }
               },
-              "required": ["data"]
```
ceorourke, line 40: Just to clarify, these are extra things that are being sent from snuba? I guess we don't care about them in sentry, just curious why it's different than https://github.com/getsentry/sentry/blob/master/src/sentry/snuba/json_schemas.py

untitaker (approved):

PR #69, opened 2023-03-28
fix(outcomes): Update outcome schema

john-z-yang:
### Overview
Remove `org_id` from outcome schema


examples/outcomes/1/outcomes5.json
```diff
@@ -0,0 +1 @@
+{"timestamp":"2023-03-28T22:51:00.000000Z","project_id":5466470,"outcome":3,"reason":"project_id","source":"pop-us","category":1,"quantity":1}
```
lynnagara: Bit of a nitpick but would you mind giving these files more useful names? e.g. This one can be `outcomes_pops.json` or whatever. So people reading these files have a better idea of what they are.

examples/outcomes/1/outcomes2.json
```diff
@@ -1,9 +1,7 @@
 {
```
lynnagara, line 1: can you rename them all? it's a little weird to have outcomes2 and 3 but not the rest of the numbers?

examples/outcomes/1/outcomes-lb.json
```diff
@@ -0,0 +1,9 @@
+{
+  "project_id": 1780659,
```
lynnagara, line 2: I know it's not exactly customer info but let's be more paranoid and please replace these all with `project_id`: 1

lynnagara: Same goes for org ids

examples/outcomes/1/outcomes-lb.json
```diff
@@ -1,9 +1,9 @@
 {
-  "project_id": 1780659,
+  "project_id": 1,
   "logging.googleapis.com/labels": {
     "host": "lb-6"
   },
-  "org_id": 0,
+  "org_id": 1,
```
lynnagara, line 6: nit: I'd leave this one as 0. Three was no org Id in this case so it filled in 0

examples/outcomes/1/outcomes-null-values.json
```diff
@@ -1,11 +1,11 @@
 {
   "timestamp": "2023-03-24T19:28:20.605851Z",
   "org_id": 1,
-  "project_id": 5,
+  "project_id": 1,
   "key_id": null,
   "outcome": 0,
   "reason": null,
-  "event_id": "1410e6d2ea534dcb9d6d15a51c9962f8",
+  "event_id": "m4CHQ8dAeS",
```
lynnagara, line 8: That's not a valid event id. Needs to be a 32 byte hex string.

lynnagara (approved):

PR #70, opened 2023-03-28
fix: Fix filename handling of jsonschema changes

untitaker:
The script is not supposed to print anything if there are no changes.

Instead, it prints all changed files right now, such as in https://github.com/getsentry/sentry-kafka-schemas/pull/69

Let's just get rid of filename headers for now and add them to the
"change object"

### DIFF
--- a/scripts/json_schema_changes.py
+++ b/scripts/json_schema_changes.py
@@ -64,12 +64,14 @@ def main() -> None:
         print_change(change)
 
     if not non_breaking_changes and not breaking_changes:
-        print("""\
+        print(
+            """\
 There were changes to the JSON schema file, but we couldn't categorize any of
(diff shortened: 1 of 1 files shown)


lynnagara (commented):
i think it may not be a bad idea to print that something changed about the schema even if the change isn't one recognized by the script. Wdyt?

lynnagara (approved):

untitaker:
@lynnagara then I would say the behavior on master is better, and we should perhaps only add a placeholder below the filename like this:

```
# schemas/some-schema.json

The file changed, but the change could not be categorized. Please fix github.com/getsentry/json-schema-diff/
```

untitaker:
^ this message should now no longer occur, but it is a nice preview i guess

lynnagara (approved):

PR #74, opened 2023-03-29
feat: Rust types

untitaker:
Add Rust type definitions using typify.

Originally I wanted to generate the source files directly into `rust/` so
people can view them with their editor directly (and more importantly that the
resulting crate would have fewer build dependencies), but `cargo publish`
actively prevents you from doing that. It has multiple checks to avoid
publishing files in gitignore, and to prevent source files from being modified
as well.

A build script is a little bit nicer than a proc_macro because the generated
files end up visible in the source code view of `cargo doc`, and because it's
possible to expose a makefile command to view the buildscript output.


schemas/ingest-metrics.v1.schema.json
```diff
@@ -22,32 +22,33 @@
...
+          "type": "object",
+          "additionalProperties": {
+            "type": "string"
+          }
         },
         "value": {
           "anyOf": [
             {
-              "type": "integer"
+              "title": "number_metric_value",
```
lynnagara, line 40: Does that also work in python?

rust/build.rs
```diff
@@ -0,0 +1,46 @@
...
+    {code}
+    }}"
+    )
+}
+
+fn main() {
+    let mut module_code = String::new();
+    module_code.push_str(&generate_schema(
+        "schemas/ingest-metrics.v1.schema.json",
+        "ingest_metrics_v1",
```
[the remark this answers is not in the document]
untitaker: i am not sure, right now i feel like it's not worth putting that much effort into the rust library to make that part happen

one annoying thing i ran into is that i would have to re-build topics parsing here because unlike in `generate_python_types.py` i actually cannot import the library before it is finished building

schemas/ingest-metrics.v1.schema.json
untitaker: I believe it will produce `NumberMetricValue = int`, but i havent' checked

rust/build.rs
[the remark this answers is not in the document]
lynnagara: Ok

lynnagara (approved):

schemas/ingest-metrics.v1.schema.json
lynnagara: I didn't see it produce anything but TypedDict's before but haven't checked recently

untitaker: I checked, and it appears it produces:

```
[...]
    value: Required[Union["NumberMetricValue", "NumbersMetricValue"]]
    """Required property"""



NumberMetricValue = Union[int, float]
"""number_metric_value."""



NumbersMetricValue = List[Union[int, float]]
"""numbers_metric_value."""
```

untitaker:
the removal of width is okay. it does not exist in prod.

PR #83, opened 2023-03-31
fix: Fix metric value types across all schemas

untitaker:
The array of numbers can only ever be an array of u32s as per Relay
sourcecode. It is used for sets


lynnagara (approved):

untitaker:
okay i'm dumb. it is definitely used for distributions as well

PR #88, opened 2023-04-01
feat: Remove unused fields from subscription results

lynnagara:
Whilst creating this schema and integrating it into Snuba and Sentry we discovered there were some fields ("profile" and "trace_output") being published that were essentially internal Snuba data and are completely unused by Sentry in the results consumer. Let's remove them from the schema and example.

Since "additionalProperties" is true, this is a backward compatible change and old messages still in the pipeline will not fail validation.

### DIFF
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -1,6 +1,3 @@
-# Search and storage is the default owner
-* @getsentry/owners-snuba
-
 # Ingest topics
 /topics/ingest-metrics.yaml                                @getsentry/owners-ingest @getsentry/owners-snuba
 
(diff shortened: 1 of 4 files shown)


wedamija (approved):
I'm not particularly against removing this, but I think extra data was included so we could log if needed for debugging. If this isn't that useful for debugging I'm happy to remove though.

lynnagara:
> I'm not particularly against removing this, but I think extra data was included so we could log if needed for debugging. If this isn't that useful for debugging I'm happy to remove though.

Thanks for the context, I did not realise it was intentional. Looking at https://github.com/getsentry/sentry/pull/46647/ I don't think it would be logged anywhere since it's immediately transformed to a `SubscriptionUpdate` without those keys... do you think we want to add 'em back there too then?

lynnagara:
Happy to close this and leave them in if you think that's best, i just assumed it was internal snuba stuff

wedamija:
> Happy to close this and leave them in if you think that's best, i just assumed it was internal snuba stuff

I don't feel too strongly either way, I'm not even sure if this info is useful for debugging. I'll leave it up to you, we can always add it back in later if we need it for some reason

lynnagara:
Since neither of us feel too strongly, I think we should merge this. Reasons:
- i don't actually see where this is being logged in sentry
- i think it's better to simplify if these fields don't really serve a purpose

PR #92, opened 2023-04-03
docs: Clarify that bumping the version is part of the release process

lynnagara:
Bumping the sentry-kafka-schemas version is part of the release process, and clarify that it is the job of the person making the release.

### DIFF
--- a/README.md
+++ b/README.md
@@ -91,6 +91,10 @@ For releasing a new stable version from main branch, go to
 [Actions](https://github.com/getsentry/sentry-kafka-schemas/actions) and
 trigger a new job for the `Release` workflow.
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -91,6 +91,10 @@ For releasing a new stable version from main branch, go to
 [Actions](https://github.com/getsentry/sentry-kafka-schemas/actions) and
 trigger a new job for the `Release` workflow.
 
+After releasing a new version, you should immediately bump Sentry, Snuba and
```
evanh, line 94: Is there an order to how these should happen? Or doesn't matter?

lynnagara: It shouldn't matter

untitaker (approved):

PR #97, opened 2023-04-06
feat(generic-metrics): Add optional use_case_id to payload schema

ayirr7:


lynnagara:
How come it's needed on the `ingest-metrics` topic? It's already in https://github.com/getsentry/sentry-kafka-schemas/blob/f4526b60b7866e6f05cc048c2a581c01036c9747/schemas/snuba-generic-metrics.v1.schema.json#L11-L13

ayirr7:
> How come it's needed on the `ingest-metrics` topic? It's already in
> 
> https://github.com/getsentry/sentry-kafka-schemas/blob/f4526b60b7866e6f05cc048c2a581c01036c9747/schemas/snuba-generic-metrics.v1.schema.json#L11-L13

The schema in that snippet is already after strings have been indexed. We're looking to add `use_case_id` to the payload before the messages are passed along to the cardinality rate limiter and to `bulk_record`

lynnagara:
Makes sense. Just out of curiosity will it always be optional? Or eventually everyone will also send it?

lynnagara (dismissed):

lynnagara:
@ayirr7 When you're ready to merge and release ping myself or @untitaker and we can go through the process together if you like.

schemas/ingest-metrics.v1.schema.json
```diff
@@ -7,6 +7,9 @@
       "title": "ingest_metric",
       "additionalProperties": false,
       "properties": {
+        "use_case_id": {
+          "type": "string"
+        },
```
lynnagara, line 12: It looks like maybe this is not going sent from relay and is only sent under very specific circumstances? Is that true? If so, I would add a description in here explaining the scenarios in which it is present, and also maybe update the name of the example file to something more descriptive? AFAIK there are no changes to Relay planned so I'm a bit surprised this would be changing since Relay is the one who writes this topic.

lynnagara (changes requested):
Confirmed offline that the messages sent from Relay will not be changing

PR #98, opened 2023-04-07
fix(events): Fix the schema for tombstone events

lynnagara:
I'm actually a bit confused by some of the other example tombstone events already checked in here because it doesn't look like Sentry is ever sending this

https://github.com/getsentry/sentry/blob/34a6dde71ea2f43c39ebc370e94692ca5e50555b/src/sentry/eventstream/snuba.py#L345-L351

### DIFF
--- a/examples/events/1/tombstone-events-no-datetime.json
+++ b/examples/events/1/tombstone-events-no-datetime.json
@@ -1,0 +1,11 @@
+[
+  2,
+  "tombstone_events",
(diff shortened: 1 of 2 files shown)


untitaker (approved):

untitaker:
I am not sure. Check the snuba code? Perhaps i was writing this based off of the code there. Not sure why I'd make it required tho

PR #109, opened 2023-04-18
feat(metrics): Add sentry_received_timestamp to schemas to support indexer SLO

ayirr7:
Supporting the work started here: https://github.com/getsentry/sentry/pull/47577. 

If we change the name of the field in the above PR, will make corresponding changes here as well.

### DIFF
--- a/examples/snuba-generic-metrics/1/snuba-generic-metrics1.json
+++ b/examples/snuba-generic-metrics/1/snuba-generic-metrics1.json
@@ -16,6 +16,7 @@
     "5": "init"
   },
   "timestamp": 1677512412,
+  "origin_timestamp": 1677519000,
   "type": "c",
   "use_case_id": "performance",
   "value": 1
--- a/examples/snuba-generic-metrics/1/snuba-generic-metrics1.json
+++ b/examples/snuba-generic-metrics/1/snuba-generic-metrics1.json
@@ -16,7 +16,7 @@
     "5": "init"
   },
   "timestamp": 1677512412,
-  "origin_timestamp": 1677519000,
+  "sentry_received_timestamp": 1677519000.456,
   "type": "c",
   "use_case_id": "performance",
   "value": 1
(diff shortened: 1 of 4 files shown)


schemas/snuba-generic-metrics.v1.schema.json
```diff
@@ -26,6 +26,9 @@
         "timestamp": {
           "type": "integer"
         },
+        "origin_timestamp": {
+          "type": "integer"
```
MeredithAnya, line 30: i think we'll need the `"number"` type here

examples/snuba-generic-metrics/1/snuba-generic-metrics1.json
```diff
@@ -16,6 +16,7 @@
     "5": "init"
   },
   "timestamp": 1677512412,
+  "origin_timestamp": 1677519000,
```
MeredithAnya, line 19: this should be a float

schemas/snuba-metrics.v1.schema.json
```diff
@@ -26,6 +26,9 @@
         "timestamp": {
           "type": "integer"
         },
+        "origin_timestamp": {
+          "type": "integer"
```
MeredithAnya, line 30: same here: "number"

schemas/snuba-metrics.v1.schema.json
```diff
@@ -66,6 +69,7 @@
         "retention_days",
         "tags",
         "timestamp",
+        "sentry_received_timestamp",
```
lynnagara, line 72: This can't be required right now, otherwise the previous message format (which is still being produced today) will fail the validation. You can make it required at a later stage once those no longer exist. Would you mind leaving one of the examples unchanged and adding a new example with the `sentry_received_timestamp`?

ayirr7: oh right oops

lynnagara (approved):

MeredithAnya (approved):

PR #113, opened 2023-04-25
ref: Add basic consumer/producer info for each topic

untitaker:
In a next step, we will use that information to give more targeted and
specific advice on how to deal with breaking changes and how to
generally roll out PRs.


lynnagara (approved):
🥳

python/sentry_kafka_schemas/sentry_kafka_schemas.py
```diff
@@ -23,19 +24,25 @@ class SchemaNotFound(Exception):
...
+class ServicesData(TypedDict):
+    consumers: Sequence[str]
+    producers: Sequence[str]
+
+
+class TopicData(TypedDict):
+    topic: str
+    description: str
+    services: ServicesData
+    schemas: Sequence[TopicSchema]
```
lynnagara, line 44: Can we hardcode the literals for the repos we support?

python/sentry_kafka_schemas/sentry_kafka_schemas.py
```diff
@@ -23,19 +24,25 @@ class SchemaNotFound(Exception):
...
+    version: int
+    type: Literal["json"]
+    compatibility_mode: Literal["none", "backward"]
+    resource: str
+    examples: Sequence[str]
+
+
+class ServicesData(TypedDict):
+    consumers: Sequence[str]
+    producers: Sequence[str]
```
lynnagara, line 37: Can we hardcode the literals for the repos we support?

untitaker: I have now hardcoded them in the test, because those type definitions are not actually enforced (we just cast to them)

I was also wondering whether we should generate the type definitions from a schema definition of topic.yml

PR #114, opened 2023-04-26
fix(events): Loosen events mechanism

lynnagara:
Same idea as https://github.com/getsentry/sentry-kafka-schemas/pull/112 but for events mechanism.

Recently we added some new properties here which caused messages to fail validation. This change removes all properties from "mechanism" except the minimal ones that Snuba consumer need and sets additionalProperties to true.

### DIFF
--- a/schemas/events.v1.schema.json
+++ b/schemas/events.v1.schema.json
@@ -701,25 +701,6 @@
         }
       ]
     },
-    "CError": {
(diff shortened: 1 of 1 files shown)


lynnagara:
Kind of confused by the bot's output today

untitaker:
so yeah.. the bot says that you loosened the schema, and right now that is a breaking change if the producer is rolled out before the consumer

if `additionalProperties` was `true` in main already, it would not be a breaking change to remove those properties

untitaker (approved):

PR #115, opened 2023-04-26
feat(geo): Add subdivision to event schema

udameli:
With subdivision is stored (https://github.com/getsentry/relay/pull/2058), this PR is updating the event schema to include it in the `geo` data.

### DIFF
--- a/schemas/events.v1.schema.json
+++ b/schemas/events.v1.schema.json
@@ -1415,6 +1415,10 @@
             "region": {
               "description": " Human readable region name or code.",
               "type": ["string", "null"]
+            },
+            "subdivision": {
+              "description": " Human readable subdivision name.",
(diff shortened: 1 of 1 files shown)


untitaker:
we should just add additionalProperties=true here if we do not index this field

udameli:
> we should just add additionalProperties=true here if we do not index this field

@untitaker here https://github.com/getsentry/sentry-kafka-schemas/blob/main/schemas/events.v1.schema.json#L1424? geo.subdivision is going to function the same way as geo.city and geo.region

untitaker:
@udameli I just saw that we do indeed access this field in snuba, so all good

udameli:
> @udameli I just saw that we do indeed access this field in snuba, so all good

thanks for checking!

PR #116, opened 2023-04-27
feat: Add support for msgpack, add schema for replay-recordings, move arroyo codecs into repo

untitaker:
# Move codecs into sentry-kafka-schemas

People are currently building their own encoding/decoding logic on top of sentry-kafka-schemas. While this was intentionally done to give people the most flexibility, it limits us in the kind of custom validation we can do on topics, and limits our ability to verify data comes in in the right serialization format.

Get rid of code like this:

```python
import random_schema_validator
import random_json_decoder
import sentry_kafka_schemas
from sentry_kafka_schemas.schema_types import FooBar

validator = random_schema_validator.compile(sentry_kafka_schemas.get_schema(...))
message = validator.validate(random_json_decoder.decode(b"..."))
message_typed = cast(FooBar, message)
```

and replace it with:

```python
import sentry_kafka_schemas
from sentry_kafka_schemas.schema_types import FooBar

codec = sentry_kafka_schemas.get_codec(...)
message_typed: FooBar = codec.decode(b"""{"hello": "world"}""")
```

We are *prescribing* `rapidjson` and `fastjsonschema` for everybody. We don’t have to get rid of the old APIs, and we can have hooks on `Codec` to swap out the libraries used. But we don’t need to build that part right now.

# Add msgpack as new “type” in topics

`true` or `{}` are json schemas that validate everything. It turns out that all jsonschema libraries accept non-json-serializable types under those schemas as well:

```python
import fastjsonschema

schema = {
    "properties": {
        "attachment_bytes": True,
    }
}

# the "true" schema accepts invalid UTF-8 bytestrings as well
fastjsonschema.compile(schema)({"attachment_bytes": b"\xc3\x28"})

fastjsonschema.compile(schema)({"attachment_bytes": 1})
```

Our Python type generation will produce `typing.Any` fields.

### Why?

- Even with msgpack as a transport layer and no type-safety for bytestrings, we do not want to give up the superior tooling that jsonschema has over other type definition languages.


lynnagara (approved):
Lgtm overall. A couple of minor things:
- Mind adding some info to the readme about msgpack and specifically how binary data should be included in the json schemas
- Shall we standardise on "description" instead of "$comment" everywhere since it's already in the other schemas?

PR #117, opened 2023-04-27
ref: Make everything take additionalProperties

untitaker:
A few of those decisions were not made deliberately, and as explained in #114 we want to generally be liberal in accepting unknown properties.

If there are individual cases where we don't want to be forward-compatible, I think we should roll those back on a case-by-case basis.


evanh (approved):

lynnagara (approved):
big_gulps_huh_2

PR #123, opened 2023-05-02
feat: JsonCodec and MsgpackCodec are usable without schemas

lynnagara:
Schema is optional in the JsonCodec and MsgpackCodec. So they can be used for topics where a schema isn't defined yet. Otherwise a separate codec will need to be defined for all topics which don't have a schema, which is a pain for users.

This is the same as the Arroyo implementation does.


untitaker (approved):

untitaker:
going to merge this so I can unblock my other PR

PR #124, opened 2023-05-03
fix: Make Codec.validate public

untitaker:
I didn't realize the entire time that that method was not on the
baseclass. It is useful because currently we are treating validation
errors as just warnings, meaning that we need to decode even if
validation fails.

We already effectively used arroyo codecs this way.

### DIFF
--- a/python/sentry_kafka_schemas/codecs/__init__.py
+++ b/python/sentry_kafka_schemas/codecs/__init__.py
@@ -30,7 +30,7 @@ class Codec(ABC, Generic[T]):
         Calling `decode(validate=False)` and then `validate()` should be
         equivalent to `decode(validate=True)`.
         """
-        raise NotImplementedError()
+        raise NotImplementedError
 
 
 class ValidationError(Exception):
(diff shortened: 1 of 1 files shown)


lynnagara (approved):
Mind raising NotImplementedError to force derived class to provide

python/sentry_kafka_schemas/codecs/__init__.py
```diff
@@ -21,6 +21,16 @@ def decode(self, raw_data: bytes, validate: bool = True) -> T:
...
 
+    @abstractmethod
+    def validate(self, data: T) -> None:
+        """
+
+        Validate a decoded Python object again.
+
+        Calling `decode(validate=False)` and then `validate()` should be
+        equivalent to `decode(validate=True)`.
+        """
```
untitaker, line 32:
```suggestion
        """
        raise NotImplementedError()
```

lynnagara:
Merging to get the dependency bumped in Sentry

PR #127, opened 2023-05-04
fix: Deduplicate list of consumers and producers

lynnagara:


untitaker (approved):
for reference, which pr exhibited this behavior?

lynnagara:
This one: https://github.com/getsentry/sentry-kafka-schemas/pull/125

PR #129, opened 2023-05-05
fix: Improve formatting of schema diff further

untitaker:
Printing a codeblock is kind of ugly -- let's try markdown lists


untitaker:
ugh 🙃 https://github.com/getsentry/action-migrations/pull/24

lynnagara (approved):

PR #135, opened 2023-05-10
fix: Some fixes to make ingest-replay-recordings usable

untitaker:


schemas/ingest-replay-recordings.v1.schema.json
```diff
@@ -2,70 +2,24 @@
...
-        "id": { "type": "string" },
-        "chunks": { "type": "integer" },
-        "size": { "type": "integer" }
-      },
-      "required": ["id", "chunks"]
-    }
-  }
+  "type": "object",
+  "properties": {
+    "type": { "const": "replay_recording_not_chunked" },
```
MeredithAnya, line 7: do not need replay chunks anymore?

untitaker: it seems that the entire chunking support was removed by replays team. there is nothing left in the consumer

untitaker:
Regarding the json-schema-diff bot: This schema is not used in prod yet. I'm just aligning it with what is in sentry codebase

PR #141, opened 2023-05-16
fix(ingest-replay-recordings): Fix consumer service

lynnagara:
I believe this is a Sentry consumer, not Snuba


untitaker (approved):
true!

PR #143, opened 2023-05-16
feat(replays): Add ingest-replay-events schema

cmanallen:
Just experimenting...

### DIFF
--- a/schemas/ingest-replay-events.v1.schema.json
+++ b/schemas/ingest-replay-events.v1.schema.json
@@ -1,0 +1,214 @@
+{
+    "$schema": "http://json-schema.org/draft-07/schema#",
+    "anyOf": [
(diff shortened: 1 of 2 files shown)


lynnagara:
@cmanallen We have a mechanism in Snuba to turn this on at a warning level - messages that fail the schema would still be processed normally but will just log a warning. If you want to merge this and try running it against prod data that works for me. Even if the schema isn't perfect, it won't cause an issue.

cmanallen:
@lynnagara Yeah we can do that.  I need to clean up the pull a bit first though.

lynnagara:
@cmanallen I'd love to get this merged - we want to make this mandatory in Snuba at some point. Do you have an example payload from this topic you can share? I can take it from there and clean up schema the if you do.

PR #145, opened 2023-05-23
doc: Generate servicemap from topic definitions

untitaker:
image


topics/snuba-queries.yaml
```diff
@@ -1,4 +1,5 @@
 topic: snuba-queries
+pipeline: snuba
```
lynnagara, line 2: This one feels like an outlier. I wonder if we should support pipeline being omitted, and in that case just render it into it's own pipeline. There are others like this (e.g. attribution topic), and I don't know that they should share the "snuba" pipeline.

lynnagara (approved):
💯

PR #149, opened 2023-05-30
feat(replays): Add version field

cmanallen:


cmanallen:
cc @untitaker

cmanallen:
> This PR contains breaking changes. 

Does it though?  Its a new field with a default value?  Surely this won't impact producer or consumer.  The field would just be ignored.

untitaker:
> Does it though? Its a new field with a default value? Surely this won't impact producer or consumer. The field would just be ignored.

We are currently considering this a breaking change because previously the schema allowed you to send any type for `version`, and now it's no longer valid to send a string for `version`.

It would not be considered a breaking change if `additionalProperties=false`.

We are still figuring this out... for now I'd suggest to ignore it. This is why it's not a required check anyway

untitaker (approved):

cmanallen:
@untitaker How long until a new tag is created?

untitaker:
@cmanallen PTAL at the bottom of the README. You can kick off a github action to start the release process

PR #150, opened 2023-06-16
feat(errors): add errors and trace.sampled to schema

barkbarkimashark:


lynnagara:
If you have an example payload to add with these fields populated it would be awesome to add to the `examples` directory. So we ensure the schema always passes correctly in CI.

schemas/events.v1.schema.json
```diff
@@ -430,6 +430,17 @@
...
+          "description": " Errors encountered during processing. Intended to be phased out in favor of\n annotation/metadata system.",
+          "anyOf": [
+            {
+              "$ref": "#/definitions/EventProcessingError"
+            },
+            {
+              "type": "null"
+            }
+          ]
+        },
```
lynnagara, line 443: Should be a list of `EventProcessingError`

lynnagara (approved):
Thanks!

PR #153, opened 2023-06-22
Add spans data schema

dbanda:
Add spans data schema. This can help us validate messages in the spans data bag.

Also:
* Fixed typo in the readme
* Add title to the top level transaction message

### DIFF
--- a/README.md
+++ b/README.md
@@ -37,7 +37,8 @@ The yaml file of a topic has 2 keys:
 ## Using the schema (in Python)
 
 ```python
-from sentry_kafka_schemas import get_codec, ValidationError
+from sentry_kafka_schemas import get_codec
+from sentry_kafka_schemas.codecs import ValidationError
 from sentry_kafka_schemas.schema_types.ingest_metrics_v1 import IngestMetric
 
 SCHEMA: Codec[IngestMetric] = get_codec("ingest-metrics")
--- a/schemas/transactions.v1.schema.json
+++ b/schemas/transactions.v1.schema.json
@@ -1,5 +1,6 @@
 {
   "$schema": "http://json-schema.org/draft-07/schema#",
+  "title": "transaction_stream_message",
   "type": "array",
   "items": [
     { "const": 2 },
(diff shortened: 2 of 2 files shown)


lynnagara:
Is there a reason why all the spans/starfish experiments has to be on the transactions topic still, and we are not creating a separate one for the new pipeline?

dbanda:
> Is there a reason why all the spans/starfish experiments has to be on the transactions topic still

@lynnagara eventually we will have spans on a new topic and ultimately spans will replace transactions. But right now, we decided to leave it there as we iterate. My understanding is starfish wants to test things and move fast. The issue is that we are stuffing a lot of things in the spans data bag and don't have any checking on it. I could create a separate schema for `SpansData`, but that would have to mean adding some customized code within the spans consumer to check the data field. Either way if fine with me, but I would like to have a schema defined somewhere so that we have a unified source of truth and catch typing bugs.

untitaker:
We have had a similar problem already. I think long-term we might want to have specialized schemas per consumer group.

>With all the schema changes that are going on, I think it's much safer to keep it separate.

The schemas are not changing. We are just parsing more/less data out of the same payloads.

What is more concerning to me is that we are trying to make sense of this data while there is absolutely zero validation for spans data in Relay. The next buggy SDK could take down a spans consumer that tries to assume anything about the payloads.

lynnagara:
My concern was coming from a place mostly unrelated to this schema change. I'm just not sure it's a good idea to build and iterate fast on things inside the existing pipeline just because it's "faster". Though if we definitely want a separate topic and schema eventually anyway, should we just go ahead and create it now?

untitaker:
> We are adding risk into our existing transactions pipeline

yes, multiple schemas per topic would solve this. there's not really a risk of one consumer group impacting another if they do not share a schema

lynnagara:
@untitaker In this case, I don't think they should even be sharing a topic at all though. My understanding is that's a shortcut we are taking right now.

untitaker:
Currently it's only sentry-kafka-schemas that makes this dangerous. I don't think there's any other reason to split out topics right now

lynnagara:
It isn't sentry-kafka-schemas that makes it dangerous though, as we are currently not enforcing the schema. I'm actually more worried about the changes we are doing in consumers and producers.

lynnagara (approved):

PR #156, opened 2023-06-23
docs: Should we recommend a tool?

lynnagara:
Many people haven't generated schemas before or don't want to write by hand. Should we recommned a tool to make the process faster?


untitaker (approved):
NGL that tool produces not great schemas, imo lots of editing is needed afterwards

PR #164, opened 2023-07-13
feat(group-attributes): add new topic and schema for group attributes

barkbarkimashark:
Define a new message type for replicating a subset of attributes from the Group, GroupAssignee, and GroupOwner postgres tables to snuba.


lynnagara (approved):
Thanks - all the examples are helpful

PR #166, opened 2023-07-19
feat: Export timer data type from querylog

lynnagara:
We use this in Snuba

### DIFF
--- a/schemas/snuba-queries.v1.schema.json
+++ b/schemas/snuba-queries.v1.schema.json
@@ -63,6 +63,7 @@
       }
     },
     "timing": {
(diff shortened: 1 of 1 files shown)


ayirr7 (approved):

ayirr7:
Just so I understand -- we're just adding this title to indicate that the schema is equivalent to the `TimerData` that we use in Snuba ?

lynnagara:
It worked in https://github.com/getsentry/snuba/pull/4529 because it just so happens to be there as `_QuerylogTiming` - we're just giving that TypedDict an explicit name

PR #169, opened 2023-07-25
feat(querylog): Add schema definition for stats object

lynnagara:
We are currently porting the querylog processor from Python to Rust, and checking the codebase to figure out what these types were supposed to be was very tricky. Let's add them to the schema.

### DIFF
--- a/schemas/snuba-queries.v1.schema.json
+++ b/schemas/snuba-queries.v1.schema.json
@@ -127,7 +127,33 @@
           "type": ["integer", "null"]
         },
         "stats": {
-          "type": "object"
+          "type": "object",
+          "properties": {
+            "final": {
+              "type": "boolean"
+            },
+            "cache_hit": {
+              "type": "integer"
+            },
+            "sample": {
+              "type": "number"
+            },
+            "max_threads": {
+              "type": "integer"
+            },
+            "clickhouse_table": {
+              "type": "string"
+            },
+            "query_id": {
+              "type": "string"
+            },
+            "is_duplicate": {
+              "type": "integer"
+            },
+            "consistent": {
+              "type": "boolean"
+            }
+          }
         },
         "status": {
           "type": "string"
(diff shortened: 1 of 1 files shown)


volokluev (commented):
Do we *have* to do this? stats is a bit of a dumping ground by design and the processor doesn't touch it. Do we need to schema-fy it?

lynnagara:
We don't need to, but porting this to rust without it is kind of a struggle. Had to read a lot of code to figure out what they were. If there's some fields that are not needed, can we clean them up and not treat querylog as a dumping ground?

schemas/snuba-queries.v1.schema.json
```diff
@@ -127,7 +127,33 @@
           "type": ["integer", "null"]
         },
         "stats": {
-          "type": "object"
+          "type": "object",
+          "properties": {
+            "final": {
+              "type": "boolean"
+            },
+            "cache_hit": {
```
volokluev, line 135: Does this mean the field is required? because I'm looking at the querylog data and it's not guaranteed to be there

schemas/snuba-queries.v1.schema.json
```diff
@@ -127,7 +127,33 @@
...
+            "max_threads": {
+              "type": "integer"
+            },
+            "clickhouse_table": {
+              "type": "string"
+            },
+            "query_id": {
+              "type": "string"
+            },
+            "is_duplicate": {
```
volokluev, line 150: same for this one

schemas/snuba-queries.v1.schema.json
lynnagara: No, default is not required unless it's explicitly added to "required"

schemas/snuba-queries.v1.schema.json
lynnagara: Same as the other one, it's not required by default

volokluev (approved):

PR #172, opened 2023-08-04
spans only schemas

dbanda:
New schema for the spans only pipeline.


examples/snuba-spans/1/basic_span.json
```diff
@@ -0,0 +1,33 @@
...
+    "parent_span_id": "deadbeefdeadbeef",
+    "segment_id": "deadbeefdeadbeef",
+    "is_segment": false,
+    "start_timestamp_ms": 1691105878720,
+    "duration_ms": 1000,
+    "exclusive_time_ms": 1000,
+    "tags": [
+      ["tag1", "value1"],
+      ["tag2", "value2"]
+    ],
```
nikhars, line 17: This should be the same format as `sentry_tags` below i.e. a map of key:value pairs. The snuba consumer should change it to make it an array of keys and values when writing it to clickhouse.

examples/snuba-spans/1/basic_span.json
```diff
@@ -0,0 +1,33 @@
+[
+  2,
```
nikhars, line 2: This version is setup when an event is sent through the eventstream. Not sure whether spans would be sent via the eventstream or some other interface.

examples/snuba-spans/1/basic_span.json
dbanda: I copied this from the way its done in transactions. Should I still switch it to a map?

examples/snuba-spans/1/basic_span.json
dbanda: removed

examples/snuba-spans/1/basic_span.json
dbanda: fixed

nikhars: I wasn't aware that transactions sends tags as arrays of keys and values. Maybe @phacops can provide better guidance on what to expect here.

nikhars (approved):

PR #174, opened 2023-09-13
fix(spans): Add missing description field

phacops:


shruthilayaj (approved):

lynnagara:
Just to confirm - new optional field?

lynnagara (approved):

phacops:
> Just to confirm - new optional field?

Not new, it's already ingested (https://github.com/getsentry/snuba/blob/master/snuba/datasets/processors/spans_processor.py#L95), I'm just making sure this schema reflects reality.

It's an optional field which can only be a `string` so we'd filter when the value is `null`, but not when the field is missing.

PR #178, opened 2023-09-26
feat(CoGS): Add `shared-resources-usage` topic schema

rahul-kumar-saini:
### Overview
- We've added a new topic and will start producing data from Snuba to it soon
- This topic is how we will share CoGS data with the Data team

### Notes
- I'm not sure if adding this here will automatically start validating the messages on this topic or if there are any side effects at all


lynnagara (approved):

schemas/shared-resources-usage.v1.schema.json
```diff
@@ -0,0 +1,35 @@
+{
+  "$schema": "http://json-schema.org/draft-06/schema#",
```
lynnagara, line 2: see the ci-failure. we enforce draft-07 so all the formats are the same

fpacifici:
Thanks a lot @rahul-kumar-saini .
@lynnagara  The usage unit is an enum actually. What do we do normally? Do we store them just as strings or do we declare enums ?

lynnagara:
No hard and fast rule but I'd go ahead an enumerate the values, I think there's no reason not to be strict here and it helps readers better understand what's expected on this topic

rahul-kumar-saini:
@lynnagara I introduced `getsentry/super-big-consumers` as a valid consumer/producer, are there any side effects of doing this? I've also adjusted the string to an enum of valid usage unit types.

lynnagara (approved):

PR #187, opened 2023-10-24
feat(feedback): add feedback as new event type

JoshFerge:
- Adds the feedback event type. See https://github.com/getsentry/team-replay/issues/222
- We'll be leveraging the issue platform to create issues, and these feedbacks will flow in from relay on the events topic and be picked up in the ingest consumer and have issues created via issue platform, which is why we need the event type on both.

### DIFF
--- a/schemas/events.v1.schema.json
+++ b/schemas/events.v1.schema.json
@@ -1077,7 +1077,7 @@
       ]
     },
     "EventType": {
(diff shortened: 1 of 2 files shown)


JoshFerge:
should be fine to roll out as this change is additive.

schemas/events.v1.schema.json
```diff
@@ -1087,7 +1087,8 @@
         "expectstaple",
         "transaction",
         "nel",
-        "default"
+        "default",
+        "feedback"
```
lynnagara, line 1091: Just out of curiosity, how come we're changing `events` too? Is it not just an issue platform change?

JoshFerge: feedbacks are going to be a new event type in addition to new issue type.

lynnagara (approved):

PR #188, opened 2023-11-02
fix: Make gauges a possible value in Snuba RH topic

ayirr7:
The schema for Snuba release health/metrics topic is not flexible enough currently to support the shared upstream ingest metrics topic schema. This adds gauges as a **possible** value to the release health topic, though it is not actually supported in release health. 

The upstream schema (ingest-metrics) is shared between BOTH generic metrics and release health, and as a result, we are noticing typing problems in Sentry codebase which we are not able to simply ignore.

### DIFF
--- a/schemas/snuba-metrics.v1.schema.json
+++ b/schemas/snuba-metrics.v1.schema.json
@@ -57,6 +57,7 @@
             },
             {
               "title": "gauge_metric_value",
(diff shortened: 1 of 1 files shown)


lynnagara (approved):

lynnagara:
@ayirr7 Should I close this PR? Or are you still intending to merge it?

ayirr7:
@lynnagara, we can close this PR. I think the original question would require a bit more thought.

PR #191, opened 2023-11-17
Embed Topics/Schemas into Rust crate

Swatinem:
When you ship a final binary using the Rust crate, that system obviously does not have all the static data available in `CARGO_MANIFEST_DIR` which is the crate path at compile time.

We thus embed all the topic metadata and schemas into the crate directly so it is available at runtime.

I also scattered a bunch of `FIXME` comments all over the place for things we can improve if we break the current API.

### DIFF
--- a/rust/lib.rs
+++ b/rust/lib.rs
@@ -1,13 +1,15 @@
+use std::path::PathBuf;
+
 use jsonschema::JSONSchema;
 use serde::{Deserialize, Serialize};
-use std::fs::{read_to_string, File};
-use std::path::{Path, PathBuf};
 use thiserror::Error;
 
 // This file is supposed to be auto-generated via rust/build.rs
 pub mod schema_types {
     include!(concat!(env!("OUT_DIR"), "/schema_types.rs"));
 }
+
+include!(concat!(env!("OUT_DIR"), "/embedded_data.rs"));
 
 #[derive(Debug, PartialEq, Serialize, Deserialize)]
 #[non_exhaustive]
@@ -36,6 +38,8 @@ pub enum SchemaError {
     InvalidType,
     #[error("Invalid schema")]
     InvalidSchema,
+    // FIXME(swatinem): maybe a dedicated `ValidationError` would be nice which
+    // carries a JSON error as well as the exact Schema error.
     #[error("Invalid message")]
     InvalidMessage,
 }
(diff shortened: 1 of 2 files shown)


rust/lib.rs
```diff
@@ -122,7 +138,7 @@ pub fn get_topic_schema(topic: &str, version: Option<u16>) -> Result<TopicSchema
     Ok(schema_metadata)
 }
 
-/// Returns the schema for a topic. If version is passed, return the schema for
+/// Returns the schema for a topic. If `version`` is passed, return the schema for
```
untitaker, line 141:
```suggestion
/// Returns the schema for a topic. If `version` is passed, return the schema for
```

rust/lib.rs
```diff
@@ -88,20 +100,24 @@ impl Schema {
...
+        let compiled = self
+            .compiled_json_schema
+            .as_ref()
+            .ok_or(SchemaError::InvalidSchema)?;
+
+        compiled
+            .validate(&message)
+            .map_err(|_| SchemaError::InvalidMessage)?;
+
+        // FIXME(swatinem): we might as well just return the `Value` here
```
untitaker, line 112: i guess we can do that but i feel like it's not useful, also i hoped to shop around for a faster json schema validator at some point and get rid of creating a Value

rust/lib.rs
```diff
@@ -68,8 +77,11 @@ pub struct Schema {
     pub version: u16,
     pub schema_type: SchemaType,
     pub compatibility_mode: CompatibilityMode,
+    // FIXME(swatinem): this should maybe be private, with an accessor, or rather &'static str?
     pub schema: String,
+    // FIXME(swatinem): this should be removed in a breaking release
```
untitaker, line 82: feel free to

rust/lib.rs
```diff
@@ -68,8 +77,11 @@ pub struct Schema {
     pub version: u16,
     pub schema_type: SchemaType,
     pub compatibility_mode: CompatibilityMode,
+    // FIXME(swatinem): this should maybe be private, with an accessor, or rather &'static str?
```
untitaker, line 80: same with version/schema_type etc. lots of boilerplate though

untitaker (approved):
i think i'm fine with breaking changes to the API if you wanna do it

rust/lib.rs
Swatinem: if we are okay with breaking the API more frequently, we can add this now and remove that later on if we find a more efficient solution. according to my benchmarks, JSON parsing is one of the most expensive parts of the consumers, and we would be doing it twice otherwise. Properly hooking this up might be a challenge though.

untitaker: either option sounds fine to me, the API does not have to be very stable, and I think it's unlikely to break any code since I believe most users won't care for this return value eitehr

untitaker (approved):

PR #195, opened 2023-12-06
feat(spans): Add a metrics summary field to spans

phacops:


lynnagara (approved):
not up to date about this specific schema... but approving since it doesn't look wrong 😀

PR #200, opened 2024-01-08
ref: Add more codeowners

untitaker:
Pierre should not be blocked on changes to the schema, he sufficiently
owns both ends of the pipe

### DIFF
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -21,16 +21,18 @@
 /topics/shared-resources-usage.yaml                        @getsentry/owners-snuba @getsentry/data
 
 # Schemas
-/schemas/profile-metadata.v1.schema.json                   @getsentry/profiling
-/schemas/profile-functions.v1.schema.json                  @getsentry/profiling
-/schemas/group-attributes.v1.schema.json                   @getsentry/issues
+/schemas/profile-metadata.v1.schema.json                   @getsentry/owners-snuba @getsentry/profiling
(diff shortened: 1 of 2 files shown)


phacops:
I would love more access like this. I would also include:
- `metrics_summaries` (`@getsentry/performance`)
- `spans` (`@getsentry/performance`)
- `profiles` (`@getsentry/profiling`)
- `functions` (`@getsentry/profiling`)

CODEOWNERS
```diff
@@ -15,7 +15,7 @@
 /topics/processed-profiles.yaml                            @getsentry/owners-snuba @getsentry/profiling
 /topics/profiles-call-tree.yaml                            @getsentry/owners-snuba @getsentry/profiling
 /topics/group-attributes.yaml                              @getsentry/owners-snuba @getsentry/issues
-/topics/snuba-spans.yaml                                   @getsentry/owners-snuba
+/topics/snuba-spans.yaml                                   @getsentry/profiling
```
lynnagara, line 18:
```suggestion
/topics/snuba-spans.yaml                                   @getsentry/owners-snuba @getsentry/profiling
```

lynnagara (changes requested):
owners-snuba should still be on this like all other snuba topics

untitaker:
it's not clear to me why we have codeowners for the topic yaml inconsistent with the examples and schema file. IMO those all should be the same

lynnagara (approved):

PR #205, opened 2024-01-19
use binary encoding

john-z-yang:


schemas/ingest-metrics.v1.schema.json
```diff
@@ -50,17 +50,12 @@
               "title": "set_metric_value",
               "type": "array",
               "items": {
-                "type": "integer",
-                "minimum": 0,
-                "maximum": 4294967295
+                "type": "number"
```
untitaker, line 53: for readability and documentation purposes, can we keep separate entries for different metric types? I understand that it doesn't make a functional difference right now, but if we want to make the schema stricter in the future (and I think it has to, because sending a string for a counter metric doesn't actually make sense), it will be easier to refactor

PR #208, opened 2024-01-23
fix(rust): Fix spurious topic-not-found due to wrong sorting key

untitaker:
In code generation, we create a sorted list of topics called TOPICS
containing key-value pairs, and then later use binary search to find an
entry by key. We don't use maps because those cannot be created at compile
time, only arrays.

Unfortunately we used a different key in each process. So we sorted by a
value like "transactions.yaml" when creating the list, but the row reads
like this:

    [
        ...
        ("transactions", "<yaml contents>"),

so the lookup happens on a value like "transactions". This is not a big
difference, but it matters because we also have
"transactions-subscription-results", which sorts after "transactions",
but before "transactions.yaml"

Refactor the build.rs so that a consistent sorting order is used, and
add a test for "tranasctions" specifically which previously would always
return "topic not found"

### DIFF
--- a/rust/build.rs
+++ b/rust/build.rs
@@ -46,34 +46,45 @@ fn generate_embedded_data() -> String {
     println!("cargo:rerun-if-changed=schemas");
 
     let manifest_root = std::env::var("CARGO_MANIFEST_DIR").unwrap_or_default();
-    let topics = enumerate_dir("topics");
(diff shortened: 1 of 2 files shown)


untitaker:
when this PR is merged we can revert https://github.com/getsentry/snuba/pull/5446

Swatinem (approved):

PR #213, opened 2024-02-13
feat(replays): add replay event to recording schema

JoshFerge:
We're now sending the replay_event to the recording consumer on an optional key. Produced in https://github.com/getsentry/relay/pull/3035, recording consumer change forthcoming


JoshFerge:
i'm a a bit confused -- adding an optional field is considered a breaking change?

lynnagara:
technically yes if additionalProperties = true, as any type would have been accepted before, and adding the property usually makes it stricter. In this case, since it's bytes and you haven't specified a type, the warning is wrong

lynnagara (approved):

PR #214, opened 2024-02-13
feat: Topic config

lynnagara:
This repo should be the source of truth for creating topics needed across various services with the correct configuration. Topics can now specify topic creation config.

This also adds all of the commit log topics and snuba dlq topics. This is needed as those topics has specific creation config that is required, and we want to have this in one place.

This replaces https://github.com/getsentry/snuba/blob/6a015671c9516af23895f04247840feaa3617c44/snuba/utils/streams/topics.py (and is where i generated these initial configs from)

Also makes the `get_topic` function public so it can be used from Snuba and other places


evanh (approved):

untitaker (approved):

python/tests/test_valid_topic_data.py
```diff
@@ -77,6 +81,7 @@ def test_all_topics() -> None:
 
     topics_dir = _TOPICS
     for filename in topics_dir.iterdir():
+        print(filename)
```
fpacifici, line 84: I guess this is not intentional

topics/group-attributes.yaml
```diff
@@ -12,3 +12,5 @@ schemas:
     resource: group-attributes.v1.schema.json
     examples:
       - group-attributes/1/
+topic_creation_config:
+  message.timestamp.type: LogAppendTime
```
fpacifici, line 16: Were these wrong or did you just want to add LogAppendfTime by default ?

CODEOWNERS
```diff
@@ -1,50 +1,71 @@
...
+/topics/outcomes.yaml                                          @getsentry/owners-snuba
+/topics/snuba-metrics.yaml                                     @getsentry/owners-snuba
+/topics/snuba-generic-metrics.yaml                             @getsentry/owners-snuba
+/topics/generic-events.yaml                                    @getsentry/owners-snuba @getsentry/issues
+/topics/ingest-replay-events.yaml                              @getsentry/owners-snuba @getsentry/replay
+/topics/processed-profiles.yaml                                @getsentry/owners-snuba @getsentry/profiling
+/topics/profiles-call-tree.yaml                                @getsentry/owners-snuba @getsentry/profiling
+/topics/group-attributes.yaml                                  @getsentry/owners-snuba @getsentry/issues
+/topics/snuba-spans.yaml                                       @getsentry/owners-snuba @getsentry/profiling
+/topics/snuba-metrics-summaries.yaml                           @getsentry/owners-snuba @getsentry/profiling
```
fpacifici, line 18: Have you ever considered laying out the repo in order to allow product teams to have their own directory that contain topic, schema and examples so that code owners do not have to be updated for each new topic ?
I'd be curious to hear if that was considered and discarded.

lynnagara: I haven't personally considered it. Though my feeling is that layout is probably good for codeowners and not a lot else. I think that layout comes with it's own challenges. You can see it in Snuba's dataset layout for example which uses something like what you mention. There I find it hard to find the right files when they are nested and it also means startup time for consumers (which only care about one storage) is pretty suboptimal as everything must always be loaded beforehand, and the code to load everything gets pretty complicated.

PR #218, opened 2024-02-15
feat: Export list_topics

lynnagara:
So it can be used to create all topics


fpacifici:
> So it can be used to create all topics

What is the goal ?

lynnagara:
can be used to bootstrap topics in different environments - e.g. self hosted / devserver could use it to bootstrap every topic. we could sync it with the values in the ops repos so make sure they're aligned. basically make sure this list stays in sync with any topics referenced elsewhere

PR #220, opened 2024-02-20
perf(gen-metrics): Remove value validation for indexer schema

john-z-yang:
### Overview

The indexer does not care about what the value field is, the validation for value should happen further down the pipeline. Updates the schema to remove value field constraints.

### DIFF
--- a/schemas/ingest-metrics.v1.schema.json
+++ b/schemas/ingest-metrics.v1.schema.json
@@ -40,53 +40,7 @@
             "type": "string"
           }
         },
-        "value": {
-          "anyOf": [
-            {
-              "title": "counter_metric_value",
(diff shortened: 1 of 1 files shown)


lynnagara (approved):

ayirr7:
This looks good to me. Overall, this just means we're skipping validation for `values` in the indexer, right? (And this also applies for release health since they share the schema)

ayirr7 (approved):

john-z-yang:
> This looks good to me. Overall, this just means we're skipping validation for `values` in the indexer, right? (And this also applies for release health since they share the schema)

Yeah we're just checking that the field exists, nothing else

PR #222, opened 2024-02-23
Add encoded series to metrics schemas

john-z-yang:
### Overview

1. Revert my previous change where I removed the metrics type restrictions.

2. Add encoded series data to metrics payload. This current iteration does not change anything functionally, it just add support for a plain array type. But in the future we want something like

```json
{
  "value": {
    "format": "zstd",
    "data": "<b64 encoded zstd goes here>"
  }
}
```


examples/snuba-generic-metrics/1/snuba-generic-metrics-dist-encoded-plain-array.json
```diff
@@ -0,0 +1,25 @@
...
+  "mapping_meta": {
+    "d": { "65560": "d:spans/duration@second" },
+    "h": {
+      "9223372036854776017": "session.status",
+      "9223372036854776010": "environment"
+    },
+    "f": { "65691": "metric_e2e_spans_dist_k_VUW93LMS" }
+  },
+  "type": "d",
+  "value": { "format": "array", "data": [0.0, 1.1, 2.2, 3.3, 4.4, 5.5] }
```
ayirr7, line 24: @Dav1dde Maybe let's just double check that this matches with what we expect Relay to output

schemas/ingest-metrics.v1.schema.json
```diff
@@ -40,7 +40,79 @@
             "type": "string"
           }
         },
-        "value": {},
+        "value": {
```
ayirr7, line 43: Is there a reason why we decided to re-fill the `value` field here?

john-z-yang: Because type checking gets really annoying, also I realized that disabling that validation won't do us much good in terms of performance because large arrays will be sent as b64 strings anyways

examples/snuba-generic-metrics/1/snuba-generic-metrics-dist-encoded-plain-array.json
Dav1dde: That's what Relay sends:

https://github.com/getsentry/relay/blob/bacbfbede5f1d57140968c202f1c544b9c1d35ce/tests/integration/test_metrics.py#L1877-L1879

schemas/ingest-metrics.v1.schema.json
```diff
@@ -40,7 +40,79 @@
...
+          "anyOf": [
+            {
+              "title": "encoded_series_metric_value",
```
Dav1dde, line 46: Should we include the `array` format name here?

I would define each possible format separately in the schema as part of the `anyOf` and not mix fields. Slightly more duplication but a lot cleaner separation. E.g. how I sketched in the epic:

```
{
  "anyOf": [
    {
      "type": "object",
      "properties": {
        "format": { "const": "array" },
        "data": { "type": "array", "items": { "type": "number" } }
      }
    },
    {
      "type": "object",
      "properties": {
        "format": { "const": "zstd" },
        "data": { "type": "string" }
      }
    },
    ... gauges, counters, legacy sets and distributions ...
  ]
}
```

schemas/ingest-metrics.v1.schema.json
```diff
@@ -40,7 +40,79 @@
...
+              "type": "object",
+              "properties": {
+                "format": {
+                  "enum": ["array"]
```
Dav1dde, line 50: This can be const (I believe that is supported in the schema)

schemas/ingest-metrics.v1.schema.json
john-z-yang: Ah I see, I might have missed that point when discussing this with you. Updating this now

schemas/ingest-metrics.v1.schema.json
john-z-yang: Done

Dav1dde (approved):

PR #224, opened 2024-02-28
feat: Add outcomes-billing topic

lynnagara:


topics/outcomes.yaml
```diff
@@ -6,7 +6,6 @@ services:
     - getsentry/sentry
     - getsentry/relay
   consumers:
-    - getsentry/sentry
```
lynnagara, line 9: I assume this was a typo, we have no consumer in sentry for outcomes

untitaker: there is one in getsentry

untitaker (approved):

topics/outcomes.yaml
lynnagara: ah ok

PR #227, opened 2024-03-01
feat(replay): Config new ingest-feedback-events topic

aliu3ntry:
https://getsentry.atlassian.net/browse/OPS-5317
https://github.com/getsentry/sentry/issues/66100

New kafka topic for user feedback events, sourced from feedback widget or crash report modal. Schema is identical to ingest-events since feedback is currently in this topic.
- Pipeline = user-feedback (new), since this product does not depend on, and is not depended on, by other topics.
- Cluster = replays, since replay team owns user feedback. We expect the volume to be too low to justify its own cluster.

Schema references (structure of the message payload):
- [SDK](https://develop.sentry.dev/sdk/event-payloads/types/#userreportv2context)
- [relay](https://github.com/getsentry/relay/blob/69c8a18ab3c359b783bf99c8c42d0135213fd463/relay-event-schema/src/protocol/contexts/user_report_v2.rs#L14)
- [relay example](https://github.com/getsentry/relay/blob/69c8a18ab3c359b783bf99c8c42d0135213fd463/tests/integration/test_feedback.py#L4)
---> logging in this fx was used to create one of the 2 examples

Also see https://github.com/getsentry/sentry/pull/66484


topics/ingest-feedbacks.yaml
```diff
@@ -0,0 +1,17 @@
+topic: ingest-feedbacks
```
lynnagara, line 1: isn't feedback the plural of feedback?

topics/ingest-feedbacks.yaml
```diff
@@ -0,0 +1,17 @@
...
+pipeline: replays
```
lynnagara, line 2: is it actually part of the replays product? isn't this just general user feedback which may or may not be associated with a replay?

topics/ingest-feedbacks.yaml
```diff
@@ -0,0 +1,17 @@
...
+description: User feedback widget events
+services:
+  producers:
+    - getsentry/sentry
+  consumers:
+    - getsentry/sentry
```
lynnagara, line 8: just out of curiosity, why does this moving to Kafka at all if Sentry is both producer and consumer and it is low volume? Our existing user feedback product is not a Kafka pipeline.

topics/ingest-feedbacks.yaml
```diff
@@ -0,0 +1,17 @@
...
+schemas:
+  - version: 1
+    compatibility_mode: none
+    type: json
+    resource: events.v1.schema.json
+    examples:
+      - events/1/
```
lynnagara, line 15: you need to provide a schema as well as examples for your new topic

topics/ingest-feedbacks.yaml
```diff
@@ -0,0 +1,17 @@
...
+topic_creation_config:
+  message.timestamp.type: LogAppendTime
```
lynnagara, line 17: I don't think you need this, you can remove the whole topic_creation_config block unless you have some special configuration you want to enforce

topics/ingest-feedbacks.yaml
aliu3ntry: yeah technically, but I think we're thinking of each event in this topic, or each user submission, as a "feedback", so the s makes this clear
@cmanallen

topics/ingest-feedbacks.yaml
aliu3ntry: made a mistake, producer is relay

topics/ingest-feedbacks.yaml
aliu3ntry: yes it's not always associated. can you elaborate on what "pipeline" is for? 

The idea was to use the same cluster/infrastructure, not cause the products are dependent, but because the same team owns them. Also to save costs

lynnagara: it provides an view on what features are linked together and how data flows through the system.

As an example, if `ingest-events` is broken, then data is not going through to `events` and other related topics as well.

It's got nothing to do with which teams are working on what at sentry at any given point it time (which unfortunately seems to change every quarter).

I think you could probably just make this `errors` here, since it's moving from the shared errors topic today.

topics/ingest-feedbacks.yaml
lynnagara: Based on Colton's comment https://github.com/getsentry/sentry/pull/66169#discussion_r1511527232
it sounds a lot like this is supposed to match the schema of `ingest-events`, not `events`

topics/ingest-feedbacks.yaml
aliu3ntry: Ok, thanks for the explanation! what does the `errors` pipeline's dependency chain look like?
I think part of the reason we want to move it off that topic is because feedback is _not_ dependent -- see motivation in https://github.com/getsentry/sentry/issues/66100. Thoughts on making a new pipeline for this topic?

Also, can we still have it on the same cluster, if it's a different pipeline? And does cluster need to be defined in some config file?

lynnagara: There's quite a few things in the errors pipeline (you can search for the files in this director that contain `errors` if you want to see some example topics). New pipeline sounds fine -- you can just change it here to `user-feedback` or whatever you like here, there's nothing else to do.

Yes, the cluster topology is independent and can always be mapped to whatever kafka cluster you like per environment later.

topics/ingest-feedbacks.yaml
lynnagara: If the schema/examples are different or going to diverge, I would suggest making a different file now, but i leave that up to you.

But what i'm specifically saying is that `events` is not the same as `ingest-events` at all. It sounds like today the schema more closely matches `ingest-events` since they are literally combined today. `events` is an entirely different thing.

aliu3ntry:
Summary of last commit:
- move to new 'user-feedback' pipeline
- add an initial schema generated from a widget feedback (captured by chrome dev tools) and https://github.com/glideapps/quicktype
- rename to 'ingest-feedback-events' since 'FeedbackEvent' seems to be the corresponding type in Django + SDK/typescript

Open to discussion on the renaming!

CODEOWNERS
```diff
@@ -2,6 +2,7 @@
 /topics/ingest-metrics.yaml                                    @getsentry/owners-ingest @getsentry/owners-snuba
 /topics/ingest-performance-metrics.yaml                        @getsentry/owners-ingest @getsentry/owners-snuba
 /topics/ingest-replay-recordings.yaml                          @getsentry/owners-ingest @getsentry/owners-snuba @getsentry/replay
+/topics/ingest-feedback-events.yaml                            @getsentry/owners-ingest @getsentry/owners-snuba @getsentry/replay
```
lynnagara, line 5: you can remove @getsentry/owners-ingest

topics/ingest-feedback-events.yaml
```diff
@@ -0,0 +1,15 @@
...
+    - getsentry/relay
+  consumers:
+    - getsentry/sentry
+schemas:
+  - version: 1
+    compatibility_mode: none
+    type: json
+    resource: ingest-feedback-events.v1.schema.json
+    examples:
+      - feedback-events/1/
```
lynnagara, line 15: just a nit but by convention, we try and align all the namings to make things easier to find here. do you mind making this folder `ingest-feedback-events` too?

examples/ingest-feedback-events/1/widget-react.json
```diff
@@ -0,0 +1,88 @@
...
+    "organization.slug": "sentry",
+    "plan": "am2_business_ent_auf",
+    "plan.name": "Business",
+    "plan.max_members": "null",
+    "plan.total_members": "323",
+    "plan.tier": "am2",
+    "timeOrigin.mode": "navigationStart"
+  },
+  "user": {
+    "ip_address": "98.51.47.219",
```
aliu3ntry, line 80: oops might have committed some of my own PII. Is there a need to revert?

examples/ingest-feedback-events/1/widget-react.json
```diff
@@ -0,0 +1,88 @@
...
+    "email": "andrew.liu@sentry.io",
```
bruno-garcia, line 81: client IPs are usually dynamic and rotate with time. It's fine to edit the PR before merging
Your email is in your git commits too btw so not a big deal

```suggestion
    "ip_address": "192.168.47.219",
    "email": "andrew.liu@example.com",
```

aliu3ntry: oh yeah and everything's squashed right, so no prob. thanks!

topics/ingest-feedback-events.yaml
```diff
@@ -0,0 +1,15 @@
...
+description: User feedback events from SDK
+services:
+  producers:
+    - getsentry/relay
+  consumers:
+    - getsentry/sentry
+schemas:
+  - version: 1
+    compatibility_mode: none
+    type: json
```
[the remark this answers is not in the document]
aliu3ntry: @JoshFerge I believe you mentioned this in our team sync? We use json, right?

topics/ingest-feedback-events.yaml
```diff
@@ -0,0 +1,15 @@
...
+    resource: ingest-feedback-events.v1.schema.json
+    examples:
+      - ingest-feedback-events/1/
```
lynnagara, line 15: One more thing -- let's add the `max.message.bytes` for this topic too? We don't want a repeat of today's INC-670 where no one is sure what the topic's max size is.

examples/ingest-feedback-events/1/widget-angular.json
```diff
@@ -0,0 +1,67 @@
+{
```
lynnagara, line 1: How come these examples don't have project_id? Isn't that a required field in user feedback? It's not on the schema either.

aliu3ntry:
Turns out relay wraps my previous schema in the msgpack 'payload' field, as an encoded JSON string. So the new schema is identical to ingest-events. Bytes is a not a valid jsonschema type so we have to omit type from 'payload'

lynnagara (approved):

PR #238, opened 2024-03-15
feat(feedback): add feedback-events DLQ

aliu3ntry:


topics/ingest-feedback-events-dlq.yaml
```diff
@@ -0,0 +1,17 @@
+topic: ingest-feedback-events-dlq
+pipeline: user-feedback
+description: DLQ for ingest-feedback-events
+services:
+  producers:
```
JoshFerge, line 5: what does producers signify?

CODEOWNERS
```diff
@@ -9,6 +9,7 @@
 /topics/ingest-events-dlq.yaml                                 @getsentry/owners-snuba
 /topics/ingest-metrics-dlq.yaml                                @getsentry/owners-snuba
 /topics/ingest-generic-metrics-dlq.yaml                        @getsentry/owners-snuba
+/topics/ingest-feedback-events-dlq.yaml                        @getsentry/owners-snuba
```
lynnagara, line 12: Isn't this replays team now?

topics/ingest-feedback-events-dlq.yaml
lynnagara: Just which services (i.e. sentry) are producing to this topic. Likely will be used soon to make these topics automatically available to sentry code via configuration.

CODEOWNERS
aliu3ntry: Oh I did this because of the pattern that all other dlq's are owners-snuba only. Would you recommend I change it to snuba+replay, or just replay?

lynnagara: Hmm, probably just replay is fine if you guys own it

lynnagara (approved):

PR #242, opened 2024-03-29
fix: Don't fail on msgpack topics

lynnagara:


rust/lib.rs
```diff
@@ -144,10 +146,6 @@ fn get_topic_schema(topic: &str, version: Option<u16>) -> Result<TopicSchema, Sc
             .ok_or(SchemaError::InvalidVersion)?
     };
 
-    if schema_metadata.schema_type != SchemaType::Json {
-        return Err(SchemaError::InvalidType);
-    }
```
fpacifici, line 149: Was that intentional ?

fpacifici (approved):

rust/lib.rs
lynnagara: Originally yeah, but it was written a long time ago before we ever supported msgpack here and we just forgot to update it.

PR #248, opened 2024-04-09
feat: Update max message size on buffered-spans

shruthilayaj:
We're hitting some max message size limits while producing to this topic. We expect this payload to be around sizes we see on transactions. Not quite bumping it to 25 MB yet (that's what transactions has), want to see how far 10 MB gets us


lynnagara:
What actual limits apply here and are being enforced in code? I don't think we can just keep bumping it up and "see if nothing breaks"

lynnagara (approved):

PR #250, opened 2024-04-10
feat: Add required partition count to topic definition

lynnagara:
There are a handful of topics at sentry that require partitions=1 and will not work properly with higher counts. Specifically the commit log topics.

Currently we only define partition counts per region in ops. This PR proposes adding the count to the topic definition where it required and enforcing this in ops (which imports this repo).

### DIFF
--- a/topics/snuba-commit-log.yaml
+++ b/topics/snuba-commit-log.yaml
@@ -18,3 +18,4 @@ topic_creation_config:
   compression.type: lz4
   cleanup.policy: compact,delete
   min.compaction.lag.ms: "3600000"
+partitions: 1  # Commit log topic must always have one partition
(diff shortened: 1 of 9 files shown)


mwarkentin (approved):
Sounds good to me, just to clarify

> This PR proposes adding the count to the topic definition where it required and enforcing this in ops (which imports this repo).

This means for these special topics only, other topics can still be set on a per-region basis?

topics/snuba-commit-log.yaml
```diff
@@ -18,3 +18,4 @@ topic_creation_config:
   compression.type: lz4
   cleanup.policy: compact,delete
   min.compaction.lag.ms: "3600000"
+partitions: 1 # Commit log topic must always have one partition
```
nikhars, line 21: Nit: It would be nicer to see the topic name and partitions together in the top of these files since partitions are an important primitive of kafka.

nikhars (approved):
Minor nit. Feel free to ignore if you want.

topics/snuba-commit-log.yaml
lynnagara: That's a fair point, decided to leave it here now though next to the topic_creation_config, as they both relate to topic creatoin

PR #251, opened 2024-04-16
Add priority to group attributes schema

snigdhas:
We're adding a new column for group_priority to the GroupAttributes table in https://github.com/getsentry/snuba/pull/5786. This updates the kafka schema to support the new field.

### DIFF
--- a/schemas/group-attributes.v1.schema.json
+++ b/schemas/group-attributes.v1.schema.json
@@ -21,6 +21,9 @@
         "substatus": {
           "type": ["integer", "null"]
         },
+        "priority": {
+          "type": ["integer", "null"]
+        },
         "first_seen": {
           "type": "string"
         },


snigdhas:
^ Not sure I understand why this change is breaking - it should be a non-required null-supported field.

vartec (approved):

schemas/group-attributes.v1.schema.json
```diff
@@ -21,6 +21,9 @@
         "substatus": {
           "type": ["integer", "null"]
         },
+        "priority": {
+          "type": ["integer", "null"]
```
wedamija, line 25: Does the null here also include not present?  Since we won't be including this column at all at first

evanpurkhiser: Not adding it to the `required` section at the bottom of the schema ensures that

wedamija (approved):

PR #258, opened 2024-04-25
feat(crons): Add monitors-clock-tick

evanpurkhiser:
We will be using this topic for https://github.com/getsentry/sentry/issues/58410


wedamija (approved):
lgtm, I guess you just need to add this to codeowners?

evanpurkhiser:
> lgtm, I guess you just need to add this to codeowners?

Ah yeah

PR #266, opened 2024-04-30
feat(bucket-encoding): Switch base64 to base64 without padding

Dav1dde:
Snuba: https://github.com/getsentry/snuba/pull/5835
Relay: https://github.com/getsentry/relay/pull/3509


Dav1dde:
Fixed with #267

PR #280, opened 2024-05-18
fix(spans): Fix precise timestamp names in example

phacops:


shruthilayaj:
@phacops let's add this schema change to buffered-segments as well so it's in sync?

phacops:
> @phacops let's add this schema change to buffered-segments as well so it's in sync?

Let's keep this PR about fixing the example that's wrong, we can make another one to sync the schema properly.

lynnagara (approved):

PR #281, opened 2024-05-22
introduce standard segment settings for all commit log topics

lynnagara:
very large segments that roll over infrequently are problematic on commit log topics because the topics will roll over infrequently causing the topics to get too large. this is problematic for post process forwarders specifically which need to scan the entire commit log on startup.

these values are taken from the events and transactions commit log topics in sentry production, we need to actually apply them across the board.

### DIFF
--- a/topics/snuba-generic-metrics-gauges-commit-log.yaml
+++ b/topics/snuba-generic-metrics-gauges-commit-log.yaml
@@ -19,3 +19,6 @@ topic_creation_config:
   cleanup.policy: compact,delete
   min.compaction.lag.ms: "3600000"
   retention.ms: "86400000"
+  segment.bytes: "104857600"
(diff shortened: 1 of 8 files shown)


vartec (approved):

topics/snuba-generic-metrics-gauges-commit-log.yaml
```diff
@@ -19,3 +19,6 @@ topic_creation_config:
   cleanup.policy: compact,delete
   min.compaction.lag.ms: "3600000"
   retention.ms: "86400000"
+  segment.bytes: "104857600"
+  segment.ms: "3600000"
+partitions: 1 # Commit log topic must always have one partition
```
lynnagara, line 24: this was missing from gauges, should be there on all commit log topics

PR #286, opened 2024-05-30
feat(issue-search): Add first_release_id to the group_attributes kafka schema

snigdhas:
Update the schema to handle the new `group_first_release_id` column added to GroupAttributes in https://github.com/getsentry/snuba/pull/5986

### DIFF
--- a/schemas/group-attributes.v1.schema.json
+++ b/schemas/group-attributes.v1.schema.json
@@ -25,7 +25,7 @@
           "type": ["integer", "null"]
         },
         "first_release_id": {
-          "$ref": "#/definitions/UUID"
+          "$ref": "#/definitions/GroupAttributesSnapshot/UUID"
         },
         "first_seen": {
           "type": "string"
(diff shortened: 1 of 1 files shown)


vartec (approved):

evanh:
@snigdhas Why is this a breaking change? Can you look into that failure?

lynnagara:
@evanh it's **technically** breaking change because `additionalProperties` was set to true before, and thus there may have been messages with this key and a different type earlier that were considered valid. But if you know were never sending `first_release_id` as something else before, you're fine.

snigdhas:
> But if you know were never sending first_release_id as something else before, you're fine

Yep, brand new column on the GroupAttributes table (added [here](https://github.com/getsentry/snuba/pull/5986)) and hasn't been included in the messages.

evanh (approved):
OK, pending the CI being fixed this is fine then.

lynnagara:
the `json-schema-diff` job is set up to fail on breaking changes as a warning but is not required, feel free to merge through that particular failure

PR #294, opened 2024-06-05
feat(uptime): Add uptime-results topic

evanpurkhiser:


CODEOWNERS
```diff
@@ -62,6 +62,9 @@
 /topics/monitors-clock-tick.yaml                               @getsentry/crons
 /topics/monitors-clock-tasks.yaml                              @getsentry/crons
 
+# Topcis related to uptime
```
lynnagara, line 65:
```suggestion
# Topics related to uptime
```

topics/uptime-results.yaml
```diff
@@ -0,0 +1,20 @@
...
+    compatibility_mode: none
+    type: json
+    resource: uptime-results.v1.schema.json
+    examples:
+      - uptime-results/1/
+topic_creation_config:
+  compression.type: lz4
+  message.timestamp.type: LogAppendTime
+  retention.ms: "86400000"
+partitions: 64
```
lynnagara, line 20: Please remove this line. It's only intended to capture topics where partitions MUST be fixed to 1. Though I see how confusing this is and I'll add some proper validation around this as a follow up.

evanpurkhiser: ah ok. Maybe we can rename it to `enforced_partition_count` or something similar 👍

lynnagara: yeah, i like that name

lynnagara (approved):

PR #299, opened 2024-06-06
ref: remove redundant property from topic files

lynnagara:
it's already specified in the filename, we can also remove the test that asserts they are the same


evanpurkhiser (approved):

evanpurkhiser:
README needs updated

fpacifici (approved):

PR #301, opened 2024-06-06
add schema, examples and topic definition for ingest-occurrences

lynnagara:
we need to add all the missing topics here so they can be used to generate topic/consumer group configuration in prod

i figured out the schema/required fields by reading the consumer code in https://github.com/getsentry/sentry/blob/671811bab8661ea5bd7c1dc95cd5bc341bf8a22d/src/sentry/issues/occurrence_consumer.py

### DIFF
--- a/schemas/ingest-occurrences.v1.schema.json
+++ b/schemas/ingest-occurrences.v1.schema.json
@@ -6,18 +6,15 @@
     "id": { "type": "string" },
     "project_id": { "type": "integer" },
     "fingerprint": { "type": "array", "items": { "type": "string" } },
-    "issue_title": { "type": "string" },
-    "subtitle": { "type": "string" },
-    "type": { "type": "integer" },
-    "initial_issue_priority": { "type": ["null", "integer"] },
-    "payload_type": { "type": "string" }
+    "issue_title": { "type": ["string", "null"] },
+    "subtitle": { "type": ["string", "null"] },
+    "type": { "type": ["integer", "null"] },
+    "initial_issue_priority": { "type": ["integer", "null"] },
+    "payload_type": {"type": ["string", "null"] }
   },
   "required": [
     "id",
     "project_id",
-    "fingerprint",
-    "issue_title",
-    "subtitle",
-    "type"
+    "fingerprint"
   ]
 }
--- a/schemas/ingest-occurrences.v1.schema.json
+++ b/schemas/ingest-occurrences.v1.schema.json
@@ -1,0 +1,15 @@
+{
+    "$schema": "http://json-schema.org/draft-07/schema#",
+    "description": "Issue occurrence",
+    "type": "object",
+    "properties": {
+        "id": {"type": "string"},
+        "project_id": {"type": "integer"},
+        "fingerprint": { "type": "array", "items": { "type": "string" } },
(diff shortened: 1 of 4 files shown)


victoria-yining-huang (approved):

lynnagara:
@snigdhas @wedamija does this look right to you?

schemas/ingest-occurrences.v1.schema.json
```diff
@@ -0,0 +1,22 @@
...
+    "type": { "type": "integer" },
+    "initial_issue_priority": { "type": ["null", "integer"] }
+  },
+  "required": [
+    "id",
+    "project_id",
+    "fingerprint",
+    "issue_title",
+    "subtitle",
+    "type"
```
snigdhas, line 20: Is this the schema for issue platform (handled [here](https://github.com/getsentry/sentry/blob/671811bab8661ea5bd7c1dc95cd5bc341bf8a22d/src/sentry/issues/occurrence_consumer.py#L332))?

If so, these fields are not required. We'll also want to add `payload_type` as another not-required field. Not too familiar with kafka schemas so I'm not sure how this handles extra fields not included in the schema

lynnagara: yes it is that consumer. are you saying `issue_title`, `subtitle` and `type` are not required?

reading that code, it seemed like these lines in the consumer would fail if they are not there:
- https://github.com/getsentry/sentry/blob/671811bab8661ea5bd7c1dc95cd5bc341bf8a22d/src/sentry/issues/occurrence_consumer.py#L172
- https://github.com/getsentry/sentry/blob/671811bab8661ea5bd7c1dc95cd5bc341bf8a22d/src/sentry/issues/occurrence_consumer.py#L167
- https://github.com/getsentry/sentry/blob/671811bab8661ea5bd7c1dc95cd5bc341bf8a22d/src/sentry/issues/occurrence_consumer.py#L168

.. as they say `payload["type"]` as opposed to `payload.get("type"`

lynnagara: updated PR with the `payload_type`

snigdhas: yeahhh this message format is a bit weird because we're essentially supporting two totally different messages based on the `payload_type` 😥

if `payload_type == occurrence | None`, it ends up at the code you mentioned, which does require those fields.
if `payload_type == status_change`, we'll process [these fields](https://github.com/getsentry/sentry/blob/671811bab8661ea5bd7c1dc95cd5bc341bf8a22d/src/sentry/issues/status_change_consumer.py#L153), which requires `new_status` but not `issue_title`, `subtitle` and `type`. 

The required fields for each path are _essentially_ mutually exclusive.

lynnagara: updated the following fields to nullable + they are also not on required list
- payload_type
- type
- issue_title
- subtitle

snigdhas (approved):
Thanks for putting up with the back and forth, lgtm!

schemas/ingest-occurrences.v1.schema.json
wedamija: Should there be sub schemas or something based on the type?

PR #303, opened 2024-06-06
fix(uptime): Allow request_info and http_status_code to be null

evanpurkhiser:


wedamija (approved):
Ahh makes sense, we might not have `request_info` for miss status, 👍

PR #307, opened 2024-06-13
add topic and schema definition for profiles topic

lynnagara:


evanh (approved):

schemas/profiles.v1.schema.json
```diff
@@ -0,0 +1,18 @@
+{
+  "$schema": "http://json-schema.org/draft-07/schema#",
+  "title": "profiles",
```
phacops, line 3: I would call that `profile`, otherwise the generated class will have a plural name.

schemas/profiles.v1.schema.json
```diff
@@ -0,0 +1,18 @@
...
+  "description": "profile data from vroom?",
```
phacops, line 4: This is an unprocessed profile message emitted from Relay.

phacops (approved):

schemas/profiles.v1.schema.json
lynnagara: oops, thanks

PR #310, opened 2024-06-18
fix(group-attributes): Update first_release_id field to integer or null

snigdhas:
This field was added to GroupAttributes as UUID but release IDs are bigints. We're updating the column in the table and the schema should be updated to match as well. 

The `first_release_id` field isn't populated anywhere yet so this is a safe change.


snigdhas:
We don't use this field anywhere yet, none of the producers include the field. Should be okay to disregard the warning ^

wedamija (approved):

PR #313, opened 2024-06-25
feat(uptime): Add uptime-results topic

evanpurkhiser:


schemas/uptime-configs.v1.schema.json
```diff
@@ -0,0 +1,37 @@
...
+          "description": "UUID of the subscription that this check config represents.",
+          "type": "string"
+        },
+        "interval": {
+          "$ref": "#/definitions/CheckInterval"
+        },
+        "timeout": {
+          "description": "The total time we will allow to make the request in seconds.",
+          "type": "number"
+        },
```
wedamija, line 28: If we go over the max will we just clamp it on the uptime checker side? Or I guess maybe we should just enforce it in sentry instead...

topics/uptime-configs.yaml
```diff
@@ -0,0 +1,28 @@
...
+  consumers:
+    - getsentry/uptime-checker
+schemas:
+  - version: 1
+    compatibility_mode: none
+    type: msgpack
+    resource: uptime-configs.v1.schema.json
+    examples:
+      - uptime-configs/1/
+topic_creation_config:
```
wedamija, line 20: Not sure if we need to set any of these other values https://docs.confluent.io/kafka/design/log_compaction.html#configure-compaction

wedamija (approved):

schemas/uptime-configs.v1.schema.json
evanpurkhiser: 👍

topics/uptime-configs.yaml
evanpurkhiser: Read through them. Doesn't seem like we need any

lynnagara (approved):

PR #318, opened 2024-06-27
fix(group-attributes): Update name for field to `first_release`

snigdhas:
Updating the name for this field to `first_release` to match the column, `GroupAttributes.group_first_release`. This field is still unused so the name change shouldn't affect any producers.


snigdhas:
^This field is still unused and not required so the name change shouldn't affect any producers.

volokluev (approved):

PR #319, opened 2024-06-27
ensure dlq topics have the same message size as their underlying topics

lynnagara:
fix `buffered-segments-dlq`, `ingest-attachments-dlq`, `ingest-metrics-dlq` and `ingest-generic-metrics-dlq`, `snuba-dead-letter-generic-metrics`, `snuba-dead-letter-querylog`, `snuba-dead-letter-replays`

also adds a test

### DIFF
--- a/python/tests/test_valid_topic_data.py
+++ b/python/tests/test_valid_topic_data.py
@@ -167,25 +167,31 @@ def test_retention() -> None:
 
 def test_dlq_configuration() -> None:
     # This file does not match the naming convention
-    custom_dlq_mapping = {"ingest-generic-metrics-dlq": "ingest-performance-metrics"}
+
+    custom_dlq_mapping = {
+        "ingest-generic-metrics-dlq": "ingest-performance-metrics",
+        "snuba-dead-letter-replays": "ingest-replay-events",
+        "snuba-dead-letter-metrics": "snuba-metrics",
+        "snuba-dead-letter-querylog": "snuba-queries",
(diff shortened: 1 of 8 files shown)


aliu39 (approved):

python/tests/test_valid_topic_data.py
```diff
@@ -163,3 +163,41 @@ def test_retention() -> None:
...
+        "snuba-dead-letter-replays": "ingest-replay-events",
+        "snuba-dead-letter-metrics": "snuba-metrics",
+        "snuba-dead-letter-querylog": "snuba-queries",
+        "snuba-dead-letter-generic-metrics": "snuba-generic-metrics",
+    }
+
+    topics_dir = _TOPICS
+    for filename in topics_dir.iterdir():
+        if filename.stem.endswith("-dlq"):
+            main_topic_name = custom_dlq_mapping.get(filename.stem, filename.stem[:-4])
```
nikhars, line 181: What does -4 represent here? Can you define a constant for it and use it instead?

Similar comment for this line as well https://github.com/getsentry/sentry-kafka-schemas/pull/319/files#diff-5977ee087c598ee3b0e4480050abfd5d9c622ab72ea34ebf9cbdaa364b397d11R184.

nikhars (approved):
Left a small comment. Though it is not a blocker.

python/tests/test_valid_topic_data.py
lynnagara: updated

PR #329, opened 2024-09-02
build(rust): Streamline dependencies

jan-auer:
Streamlines both build and runtime dependencies of the Rust crate:

- Update `jsonschema` to `0.18` and removes default features. Most notably, this
  disables the `resolve-http`, which bundles `reqwest`.
- Remove `regress`, as this doesn't seem to be referenced by generated code
  anymore.
- Remove `jsonschema` and `thiserror` from build dependencies, as these aren't
  used.

Ideally, I would also remove the duplicated build-dependencies from dependencies
as these are all enabled by **default**. However, I wasn't sure how the build
script was intended to be used, so I left it alone.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -32,14 +32,13 @@ required-features = ["type_generation"]
 
 [features]
 default = ["type_generation"]
-type_generation = ["prettyplease", "schemars", "syn", "typify", "regress"]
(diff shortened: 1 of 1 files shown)


untitaker (approved):
this usecase of "build.rs can be run standalone" might be worth making a separate crate for, then there's no issue with dependencies anymore

PR #330, opened 2024-09-13
feat(uptime): Add request{method,headers,body} fields

evanpurkhiser:
Adds to the uptime-configs topic schema the ability to specify method, headers, and body

### DIFF
--- a/schemas/uptime-configs.v1.schema.json
+++ b/schemas/uptime-configs.v1.schema.json
@@ -33,15 +33,7 @@
         "request_method": {
           "description": "The HTTP method to use for the request.",
           "type": "string",
-          "enum": [
-            "GET",
-            "POST",
-            "HEAD",
-            "PUT",
-            "DELETE",
-            "PATCH",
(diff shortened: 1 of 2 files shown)


schemas/uptime-configs.v1.schema.json
```diff
@@ -29,6 +29,22 @@
...
+        "request_headers": {
+          "description": "Additional HTTP headers to send with the request",
+          "type": "object",
+          "additionalProperties": {
+            "type": "string"
+          }
+        },
+        "request_body": {
+          "description": "Additional HTTP headers to send with the request",
+          "type": "string"
```
wedamija, line 47: Now that I'm thinking about it... should the body also just be a json dict? Mostly people are sending specific key/val params here right?

evanpurkhiser: no I think it's very reasonable someone might want to pass xml or some other format

wedamija (approved):

PR #332, opened 2024-09-16
chore(github): Bump `upload-artifact` to v4

wedamija:
The release action is now failing, seeing if just bumping this fixes it

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -18,7 +18,7 @@ jobs:
         with:
           python-version: 3.11
       - run: make build
-      - uses: actions/upload-artifact@v2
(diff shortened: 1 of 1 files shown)


hubertdeng123 (approved):

.github/workflows/build.yml
```diff
@@ -18,7 +18,7 @@ jobs:
         with:
           python-version: 3.11
       - run: make build
-      - uses: actions/upload-artifact@v2
+      - uses: actions/upload-artifact@v4
         if: github.event_name != 'pull_request'
         with:
           name: ${{ github.sha }}
```
asottile-sentry, line 24: an aside: this should happen *after* validation in case it's actually broken

PR #335, opened 2024-09-25
ref(eap): Update interface of eap mutations

untitaker:
https://github.com/getsentry/snuba/pull/6344

### DIFF
--- a/examples/snuba-eap-mutations/mutation-1.json
+++ b/examples/snuba-eap-mutations/mutation-1.json
@@ -1,7 +1,7 @@
 {
   "filter": {
     "organization_id": 1500,
-    "_sort_timestamp": 150,
+    "start_timestamp_ms": 150,
     "trace_id": "deadbeefdeadbeefdeadbeefdeadbeef",
     "span_id": "deadbeefdeadbeef"
   },
--- a/schemas/snuba-eap-mutations.v1.schema.json
+++ b/schemas/snuba-eap-mutations.v1.schema.json
@@ -13,8 +13,9 @@
             "organization_id": {
               "$ref": "#/definitions/UInt"
             },
-            "start_timestamp_ms": {
-              "$ref": "#/definitions/UInt"
+            "start_timestamp": {
+              "description": "Corresponds to start_timestamp_precise in snuba-spans schema",
+              "$ref": "#/definitions/PositiveFloat"
             },
             "trace_id": {
               "$ref": "#/definitions/UUID",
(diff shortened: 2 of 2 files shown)


examples/snuba-eap-mutations/mutation-1.json
```diff
@@ -1,7 +1,7 @@
 {
   "filter": {
     "organization_id": 1500,
-    "_sort_timestamp": 150,
```
ayirr7, line 4: Curious why we had to make this change?

untitaker: not strictly necessary, but now there is just one place that computes _sort_timestamp (the snuba consumers), instead of whichever users of the mutations platform we have. `_sort_timestamp` is an attribute derived from `start_timestamp_ms` and IMO should've never been exposed outside of snuba in the first place

untitaker: please check https://github.com/getsentry/snuba/pull/6344/files

ayirr7 (approved):

examples/snuba-eap-mutations/mutation-1.json
evanh: I think we need to be a little careful here: `_sort_timestamp` exists so to abstract which timestamp values we use to sort the data. So it might not always be start timestamp.

evanh (approved):

examples/snuba-eap-mutations/mutation-1.json
untitaker: @evanh I think we have to either expose _sort_timestamp or start_timestamp_ms to users of the mutation interface. But there's no rush to make this interface change.

phacops: We should only expose `start_timestamp` to users of the mutation interface and note that it's meant to be a microsecond precision timestamp, not milliseconds.

untitaker: this is how start timestamp is represented in a span on kafka: https://github.com/getsentry/snuba/blob/a612e70b017a95969a6f6f21332952cf5d386b21/rust_snuba/src/processors/eap_spans.rs#L300

intended to match that

what is the interface supposed to be for EAP then?

phacops: > I think we need to be a little careful here: `_sort_timestamp` exists so to abstract which timestamp values we use to sort the data. So it might not always be start timestamp.

While it's true, I think there's a higher chance we move towards sending incomplete spans from the SDKs and having to update `end_timestamp`, which makes it not viable for the primary key.

phacops:
This experiment was shut down.

PR #336, opened 2024-10-02
Changes the schema so Querylog can use `progress_bytes` instead of `bytes`

xurui-c:
unblocks https://github.com/getsentry/snuba/pull/6371

### DIFF
--- a/examples/snuba-queries/1/snuba-queries-empty-trace-id.json
+++ b/examples/snuba-queries/1/snuba-queries-empty-trace-id.json
@@ -64,6 +64,7 @@
       },
       "result_profile": {
         "bytes": 1305,
+        "progress_bytes": 0,
         "blocks": 1,
         "rows": 22,
         "elapsed": 0.009863138198852539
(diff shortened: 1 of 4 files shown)


examples/snuba-queries/1/snuba-queries-empty-trace-id.json
```diff
@@ -63,7 +63,7 @@
         "array_join_cols": ["errors_local.tags.key", "errors_local.tags.value"]
       },
       "result_profile": {
-        "bytes": 1305,
+        "progress_bytes": 1305,
```
volokluev, line 66: just put both in, it sends both right?

schemas/snuba-queries.v1.schema.json
```diff
@@ -171,7 +171,7 @@
         "result_profile": {
           "type": ["object", "null"],
           "properties": {
-            "bytes": {
+            "progress_bytes": {
```
volokluev, line 174: don't change, add

onewland:
Is nothing referring to the old field?

Generally when renaming/replacing a field you go through a sequence like:
1. starting state (field `old` exists)
2. add field new (fields `old`, `new` exist)
3. remove field old (only field `new` exists)

That said, this is fine if you don't need that intermediate step

john-z-yang (approved):

PR #338, opened 2024-10-17
ref(snuba-bulk-deletes): add snuba-lw-deletions topic/schema

MeredithAnya:
We have been working on building out deletes in the snuba codebase, currently it's just an api that goes through a delete pipeline, somewhat similar to a regular SELECT query, but it's a DELETE. In order to be able to scale deletes, we want to be able to batch these delete requests together. So we want to add a kafka consumer to do so. 

The snuba API will still exists, but instead of running the query immediately, we will produce a message to the `snuba-lw-deletions` topic, and have a consumer that will do the work of batching, and executing the delete on the batched message (basically formatting multiple deletes into one delete query)


examples/snuba-lw-deletions/1/snuba-lw-deletions-search-issues.json
```diff
@@ -0,0 +1,13 @@
+{
+  "storage_name": "search_issues",
+  "rows_to_delete": 1000,
+  "conditions": {
+    "project_id": [1],
+    "group_id": [2222222, 3333333, 4444444]
+  },
```
kylemumma, line 7: with this input format the consumer will be doing the conversion to SQL.

An alternative design would be having the producer do the conversion, which would mean less work for the consumer.

im not sure if one approach should be preferred over the other.

schemas/snuba-lw-deletions.v1.schema.json
```diff
@@ -0,0 +1,41 @@
...
+        },
+        "group_id": {
+          "type": "array",
+          "items": { "type": "integer" }
+        }
+      },
+      "additionalProperties": false
+    }
+  }
+}
```
kylemumma, line 41: this now means that `project_id` and `group_id` being required for a delete is enforce (and would need to be changed) in at least 2 places now instead of 1. at the api schema and the topic definition.

i think its fine, but it makes me think sql query might be better for the topic than this object.

schemas/snuba-lw-deletions.v1.schema.json
```diff
@@ -0,0 +1,41 @@
...
+  "properties": {
+    "storage_name": {
+      "type": "string"
+    },
+    "rows_to_delete": {
+      "type": "integer"
+    },
+    "conditions": {
+      "type": "object",
+      "oneOf": [{ "$ref": "#/definitions/SearchIssuesConditions" }]
```
kylemumma, line 14: why do we have oneOf, instead of the ref directly?

kylemumma (approved):

examples/snuba-lw-deletions/1/snuba-lw-deletions-search-issues.json
onkar: > An alternative design would be having the producer do the conversion, which would mean less work for the consumer.

AFAIK, this format is valid only for the search issues and can change if/when we onboard other products on lightweight delete. Whichever approach we choose, let's make sure it remains consistent. I believe if input format changes from one product to another, then we may not have one place (either producer or consumer) to do the conversion. But this is a fair point that needs to be given some thought.

examples/snuba-lw-deletions/1/snuba-lw-deletions-search-issues.json
```diff
@@ -0,0 +1,13 @@
+{
+  "storage_name": "search_issues",
+  "rows_to_delete": 1000,
```
onkar, line 3: Is snuba going to honor the value of `"rows_to_delete"` that is passed in here? I thought producer-consumer will determine the batch size.

kylemumma: I believe this field represents "how many rows does this delete effect". (ex. 1000)
The consumer will use this to increment the overall batch size.

schemas/snuba-lw-deletions.v1.schema.json
MeredithAnya: so that later you can just add another definition for a different storage

examples/snuba-lw-deletions/1/snuba-lw-deletions-search-issues.json
MeredithAnya: > with this input format the consumer will be doing the conversion to SQL.

it does, but formatting the conditions uses the same logic from the api endpoint, so we can re-use that functionality.  it also makes it easier to validate the kafka message. If you had the SQL string as a parameter and you wanted to do that validation in the consumer you'd have to parse it anyway (or at least use a regex)

schemas/snuba-lw-deletions.v1.schema.json
MeredithAnya: I think its good to have the validation in the kafka topic, in case we ever make the change to produce messages to the topic from a different place than the api we have. Additionally, the validation we have in the API is to make sure that we can respond to the user if they have the wrong column/column type etc, whereas this validation is to make sure our logic doesn't make a mistake when producing messages, and that error would be seen in the consumer not the api. So I think both are needed

examples/snuba-lw-deletions/1/snuba-lw-deletions-search-issues.json
MeredithAnya: yeah this attribute is going to be used to batch up messages - though i've been thinking that maybe it should be part of the headers for the message

CODEOWNERS
```diff
@@ -109,6 +109,7 @@
 # Internal Snuba topics
 /topics/snuba-queries.yaml                                     @getsentry/owners-snuba
 /topics/event-replacements.yaml                                @getsentry/owners-snuba
+/topics/snuba-lw-deletions.yaml                                @getsentry/owners-snuba
```
lynnagara, line 112: are these deletions supposed to be for a specific storage? i'd suggest putting that in the name of the topic to make it clear to everyone + if you may have multiple deletions topics on different datasets in the future

examples/snuba-lw-deletions/1/snuba-lw-deletions-search-issues.json
lynnagara: in general, we don't really recommend putting important information in the headers of the message. i think it's more error prone, not covered by the schema, and needs to be manually serialized/deserialized. it also doesn't work with the snuba backend in dev that bypasses kafka

CODEOWNERS
MeredithAnya: @lynnagara I was planning on using one topic for all storages and have the consumer group determine which messages to process (based on the storage, which is why I was asking about the kafka headers in slack)

lynnagara: i see, please use separate topic instead. we don't have shared topics across pipelines, and i don't think we should start here. topics are cheap and easy to create. it will be more efficient for your consumers anyway to not have to download and manually handle all the messages intended for other pipelines.

lynnagara: you want to have a single consumer subscribe to all the topics and do every type of deletions?

we don't really do this today except in the post process forwarder, but it's fine to change, and we probably should anyway to align to the kafka consumer apis which all support this

PR #340, opened 2024-10-23
feat(crons): Add `mark_unknown` clock task

evanpurkhiser:
This task will be triggered when we detect an anomaly in check-in
volume during a clock tick. When this happens we are unable to know that
all check-ins before that tick have been received and will need to mark
all in-progress monitors as resulting in a 'unknown' state.

Part of https://github.com/getsentry/sentry/issues/79328

### DIFF
--- a/examples/monitors-clock-tasks/1/mark_unknown.json
+++ b/examples/monitors-clock-tasks/1/mark_unknown.json
@@ -1,0 +1,6 @@
+{
+  "type": "mark_unknown",
+  "ts": 1714072962,
+  "monitor_environment_id": 1,
+  "checkin_id": 1
+}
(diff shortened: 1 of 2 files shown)


examples/monitors-clock-tasks/1/mark_unknown.json
```diff
@@ -0,0 +1,6 @@
+{
+  "type": "mark_unknown",
+  "ts": 1714072962,
+  "monitor_environment_id": 1,
+  "checkin_id": 1
+}
```
wedamija, line 6: I thought that when we detected an outage, we'd just set a variable somewhere so that when we go to mark a miss, we mark an unknown instead?

evanpurkhiser: Yeah we will. This is for marking all in progress as unknown status

wedamija (approved):

PR #343, opened 2024-10-28
feat: Add taskworker schema

markstory:
The changes in getsentry/sentry#79742 are blocked because the `task-worker` topic does not have schema metadata.

While I would like this topic to contain protobuf messages, we lack tooling for that here at this time. As a temporary solution, I've added JSON schema for the task-worker topic, and will work on adding protobuf support separately.

### DIFF
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -74,6 +74,8 @@
 /topics/uptime-results.yaml                                    @getsentry/crons
 /topics/uptime-configs.yaml                                    @getsentry/crons
 
+/topics/task-worker.yaml                                       @getsentry/hybrid-cloud
+
 # Schemas
 /schemas/profile-metadata.v1.schema.json                       @getsentry/profiling
 /schemas/profile-functions.v1.schema.json                      @getsentry/profiling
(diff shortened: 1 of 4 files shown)


CODEOWNERS
```diff
@@ -105,6 +109,7 @@
 /examples/monitors-clock-tasks/                                @getsentry/crons
 /examples/uptime-results/                                      @getsentry/crons
 /examples/uptime-configs/                                      @getsentry/crons
+/examples/taskworker/                                          @getsentry/hybrid-cloud
```
markstory, line 112: This team isn't the final assignment. I need to get another team setup still.

topics/task-worker.yaml
```diff
@@ -0,0 +1,25 @@
...
+  producers:
+    - getsentry/sentry
+  consumers:
+    - getsentry/sentry
+schemas:
+  - version: 1
+    compatibility_mode: none
+    # TODO(mark) protobuf support
+    type: json
+    resource: taskworker.v1.schema.json
```
markstory, line 14: Ideally this would be `type: protobuf` and we would have a way to link to the protobuf IDL/types.

CODEOWNERS
markstory: Making a new team in getsentry/security-as-code#1061

enochtangg (approved):

PR #344, opened 2024-10-29
feat(protobuf) Add support for protobuf message topics

markstory:
For taskworkers (and likely other topics) in the future, we would like to put protobuf serialized messages into the topic instead of json/msgpack. Using protobufs will let us have one message schema for both kafka, rpc and application logic.

In order to get access to the protobuf generated code, I've needed to expand the dependencies of this library. I thought this was the simplest to build and maintain solution.


topics/task-worker.yaml
```diff
@@ -8,10 +8,9 @@ services:
     - getsentry/sentry
 schemas:
   - version: 1
-    compatibility_mode: none
-    # TODO(mark) protobuf support
-    type: json
-    resource: taskworker.v1.schema.json
+    compatibility_mode: backward
```
volokluev, line 11: how does this compatibility mode actually affect functionality?

markstory: I had interpreted this as indicating that the messages in the topic were backwards compatible. However, we have no topics with this mode right now, which makes me think it may not be complete.

evanh (approved):
This looks good to me.

The one question I had is whether we still want to maintain a separate `sentry-protos` repo now that this repo supports protobuf. We could consolidate the two.

PR #346, opened 2024-11-04
chore(eap-spans): Add a field to suggest consumers to ingest spans in EAP

phacops:


phacops:
Made the field a Kafka header instead.

PR #354, opened 2024-12-10
chore: New topic test-kraft

untitaker:


untitaker:
decided to only add a physical topic, since this won't be used by any service

PR #356, opened 2024-12-17
feat(uptime): Add changes to support execution windows

wedamija:
As part of the multi region project we want to support execution windows, so that different regions can be responsible for different segments of the overall interval for the uptime monitor.

Implements https://github.com/getsentry/uptime-checker/issues/184

### DIFF
--- a/schemas/uptime-configs.v1.schema.json
+++ b/schemas/uptime-configs.v1.schema.json
@@ -5,9 +5,9 @@
   "definitions": {
     "CheckInterval": {
       "title": "check_interval",
-      "description": "The interval between each check run in seconds.",
+      "description": "The interval between each check run in seconds. In the uptime monitors themselves we support intervals of 60, 300, 1200, 1800 and 3600. Other values in the list are for support of multi region. We'll need to support larger values than an hour at some point, but that requires changes to the scheduler in uptime-checker",
       "type": "number",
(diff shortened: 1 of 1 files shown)


schemas/uptime-configs.v1.schema.json
```diff
@@ -5,9 +5,9 @@
   "definitions": {
     "CheckInterval": {
       "title": "check_interval",
-      "description": "The interval between each check run in seconds.",
+      "description": "The interval between each check run in seconds. In the uptime monitors themselves we support intervals of 60, 300, 1200, 1800 and 3600. Other values in the list are for support of multi region. We'll need to support larger values than an hour at some point, but that requires changes to the scheduler in uptime-checker",
       "type": "number",
-      "enum": [60, 300, 600, 1200, 1800, 3600]
+      "enum": [60, 120, 180, 300, 600, 900, 1200, 1800, 2400, 3600]
```
wedamija, line 10: This is a temporary hack for now. uptime-checker schedules checks regularly within the hour, and so we can only support execution windows that happen at a regular place.

We probably need to update the scheduler to instead choose an epoch, then based on the different between the epoch and the current tick calculate what needs to run next. We might need to switch to a priority queue to make this simpler.

@evanpurkhiser I vaguely remember there being some reason the priority queue was worse than what we implemented, do you remember why?

PR #359, opened 2024-12-20
feat(uptime): new snuba-uptime-results topic

JoshFerge:
creates a new kafka topic `snuba-uptime-results` which will be used for sending uptime results to snuba.

needed for https://github.com/getsentry/team-uptime/issues/23


schemas/snuba-uptime-results.v1.schema.json
```diff
@@ -0,0 +1,145 @@
+{
+  "$schema": "http://json-schema.org/draft-07/schema#",
+  "title": "snuba_uptime_result",
+  "definitions": {
```
JoshFerge, line 4: these are copied from https://github.com/getsentry/sentry-kafka-schemas/blob/6702856230dfaf3b15210cf2894cbb5106950ca3/schemas/uptime-results.v1.schema.json#L5

i tried to link directly to them, but it doesn't seem like our framework supports that.

topics/snuba-uptime-results.yaml
```diff
@@ -0,0 +1,18 @@
...
+services:
+  producers:
+    - getsentry/sentry
+  consumers:
+    - getsentry/snuba
+schemas:
+  - version: 1
+    compatibility_mode: backward
+    type: json
+    resource: snuba-uptime-results.v1.schema.json
```
wedamija, line 12: I think the schemas are identical, so can we just point to `uptime-results.yaml` here and avoid the duplication?

JoshFerge: they aren't identical, as the new schema contains organization_id, project_id, retention_days, and region_slug which the old one does not contain.

we could alternatively update the old schema to contain these values i suppose.

CODEOWNERS
```diff
@@ -88,6 +89,7 @@
 /schemas/monitors-clock-tick.v1.schema.json                    @getsentry/crons
 /schemas/monitors-clock-tasks.v1.schema.json                   @getsentry/crons
 /schemas/uptime-results.v1.schema.json                         @getsentry/crons
+/schemas/uptime-results.v1.schema.json                         @getsentry/crons
```
evanpurkhiser, line 92: Duplicate?

schemas/snuba-uptime-results.v1.schema.json
evanpurkhiser: @lynnagara do you know if it's possible to cross reference schemas somehow in this repo?

Would be nice to not need to duplicate these

topics/snuba-uptime-results.yaml
evanpurkhiser: 👎 to having the other schema include these values. They won't be used since the uptime checker isn't aware of org / project_id (and can't be since the requests are de-duplicated right now)

schemas/snuba-uptime-results.v1.schema.json
untitaker: @wedamija is dealing with this in #360

evanpurkhiser: Oh cool, I totally mised it. Thanks!

wedamija:
I handled avoiding the duplication here, after some changes to our schema library: https://github.com/getsentry/sentry-kafka-schemas/pull/361

PR #360, opened 2024-12-21
feat: Allow schemas to reference other schemas

wedamija:
This allows us to reference external schemas to avoid duplication. They can be referenced like
```
"$ref": "file://uptime-results.v1.schema.json#/definitions/CheckResult"
```

The handler assumes all schemas will be located in out `schemas` directory.

Usage example here: https://github.com/getsentry/sentry-kafka-schemas/pull/361

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -47,6 +47,7 @@ prettyplease = { version = "0.2.4", optional = true }
 schemars = { version = "0.8.0", optional = true }
 syn = { version = "2.0.11", optional = true }
 typify = { version = "0.0.16", optional = true }
+url = "2.5.4"
 
 [build-dependencies]
 prettyplease = { version = "0.2.4", optional = true }
--- a/python/generate_python_types.py
+++ b/python/generate_python_types.py
@@ -115,7 +115,7 @@ class FileRefResolver(RefResolver):
         for part in parts:
             if not isinstance(current, dict) or part not in current:
                 raise UnRedolvedException(f"Reference not found: {ref_path}")
-            current = current[part]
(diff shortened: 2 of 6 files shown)


wedamija:
I think that we likely need to do something to support this in rust as well - the tests seem less comprehensive there, so although they passed I doubt this works. Any thoughts on this @untitaker?

untitaker (commented):
how does `get_schema` behave? I think it will have to resolve all file refs and return a self-contained schema. it is used in consumers to validate.

wedamija:
> (for both rust/python) how does `get_schema` behave? I think it will have to resolve all file refs and return a self-contained schema. it is used in consumers to validate.

Python works after my changes - https://github.com/getsentry/sentry-kafka-schemas/pull/361 has an example. I manually tested dropping fields from the examples and it validates successfully. I'll try get a test going in rust

untitaker:
@wedamija my bad, yeah i think if we can get some tests in rust that load all schemas like we have in python we'd be golden.

wedamija:
> @wedamija my bad, yeah i think if we can get some tests in rust that load all schemas like we have in python we'd be golden.

Ok, this is tested via the schema in https://github.com/getsentry/sentry-kafka-schemas/pull/361. I can collapse these prs together if you prefer

wedamija:
> > @wedamija my bad, yeah i think if we can get some tests in rust that load all schemas like we have in python we'd be golden.
> 
> Ok, this is tested via the schema in #361. I can collapse these prs together if you prefer

Hmm, looks like I missed something in the schema generation...

untitaker (approved):

PR #364, opened 2025-01-02
Add feature flag context

cmanallen:
Ref: https://github.com/getsentry/team-replay/issues/520

### DIFF
--- a/schemas/events.v1.schema.json
+++ b/schemas/events.v1.schema.json
@@ -268,6 +268,16 @@
     "Contexts": {
       "type": "object",
       "properties": {
+        "flags": {
+          "anyOf": [
+            {
+              "$ref": "#/definitions/FeatureContext"
+            },
+            {
+              "type": "null"
+            }
+          ]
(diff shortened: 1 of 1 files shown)


cmanallen:
@untitaker Is this repo still relevant?

untitaker (approved):

untitaker:
yep. sorry for the slow review here. you can add the replays team as codeowner so you can unblock yourself (most other datasets have that)

cmanallen:
@untitaker This is on the errors schema but point received 👍.  Is the backwards compatibility error an issue for the events schema?  Not sure how to fix it.

untitaker:
@cmanallen yeah i think adding replays as codeowners to the errors schema is fine, doesn't imply exclusive ownership here

i think the backwards compat issue is OK for as long as relay will add restrictions to guarantee that the context conforms to this format

cmanallen:
@untitaker Yep.  Relay will mandate this format.

PR #365, opened 2025-01-06
add ingest-transaction-blq topic

victoria-yining-huang:


topics/ingest-transactions-blq.yaml
```diff
@@ -0,0 +1,17 @@
...
+description: Backlog queue for ingest-transactions
+services:
+  producers:
+    - getsentry/sentry
+  consumers: []
+schemas:
+  - version: 1
+    compatibility_mode: none
+    type: json
+    resource: any.json
```
lynnagara, line 11: schema should match the transactions topic, not "any"

lynnagara (commented):
1. the schema and configuration of the topic should be the same as ingest-transactions
2. `blq` is a pretty unintuitive name for me - i don't think it's a self-explanatory acronym. can we call it backlogged or something?

victoria-yining-huang:
@lynnagara addressed your comments

PR #370, opened 2025-01-20
Improve support for protocol buffer topics in rust library

markstory:
When adding support for protocol buffer support for topic schema in #344 I only implemented python support. These changes add similar support for the rust library.

Because we effectively need to look up protobuf generated code at runtime, I expanded the scope of `build.rs` to generate `PROTOS`, which is similar to `EXAMPLES` and `TOPICS`. In order to appease the compiler, I needed to use `Box<dyn Any>` and `.downcast` into the specific structure that the consuming application needs. Obviously this parse/validate logic requires having `sentry-protos` available expanding the dependencies of this project, which could spill into consuming applications as well.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -116,6 +116,8 @@ jobs:
     steps:
       - uses: actions/checkout@v2
         name: Checkout code
+      - name: Install protoc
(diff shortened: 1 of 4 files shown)


evanh (approved):
LGTM. Not sure if there is a cleaner way to do this.

PR #371, opened 2025-01-22
[WIP] chore(sourcemaps): Update stackframe kafka schema

jangjodi:
Update StackFrame kafka schema to include `data.symbolicated`
[this](https://github.com/getsentry/sentry/blob/7b0d88de1d85955cb9971392be9986ddd05f8ee1/src/sentry/lang/javascript/processing.py#L55-L64) is the data from the sentry side

### DIFF
--- a/schemas/events.v1.schema.json
+++ b/schemas/events.v1.schema.json
@@ -622,6 +622,13 @@
         }
       }
     },
+    "StackFrameMetadata": {
+      "default": null,
+      "type": ["object", "null"],
+      "additionalProperties": {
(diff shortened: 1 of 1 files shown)


yuvmen:
This PR was previoulsy opened by Jodi who has since left, we ended up not carrying the symbolicated value on each frame but instead have a top level flag for the information we want. We calculate this flag in sentry and so on the Kafka message only this flag will be carried onward, this is done in this PR instead:
https://github.com/getsentry/sentry-kafka-schemas/pull/396

PR #372, opened 2025-01-22
fill out type for MonitorCheckIn.payload

asottile-sentry:


asottile-sentry:
to fix these mypy errors:
```
src/sentry/monitors/consumers/monitor_consumer.py:984: error: Argument 1 to "loads" has incompatible type "str | int | float | dict[str, Any] | list[Any] | None"; expected "str | bytes"  [arg-type]
src/sentry/monitors/consumers/monitor_consumer.py:1002: error: Generator has incompatible item type "datetime | None"; expected "datetime"  [misc]
src/sentry/monitors/consumers/monitor_consumer.py:1036: error: Argument 1 to "loads" has incompatible type "str | int | float | dict[str, Any] | list[Any] | None"; expected "str | bytes"  [arg-type]
```

JoshFerge (approved):

asottile-sentry:
apparently we send along bytes in msgpack -- which results in something which is impossible to validate with jsonschema (since `string` doesn't allow `bytes` which makes sense)

going to see if I can convert in relay to string and then revert the revert after reverting this

PR #380, opened 2025-02-07
Creating the events backlog topic

kneeyo1:
This creates the errors/events backlog topic for kafka router. This does so in all regions.

### DIFF
--- a/topics/ingest-events-backlog.yaml
+++ b/topics/ingest-events-backlog.yaml
@@ -1,0 +1,17 @@
+Pipeline: errors
+description: Backlog queue for ingest-events
+services:
+  producers:
+    - getsentry/sentry
+  consumers: []
+schemas:
+  - version: 1
+    compatibility_mode: none
+    type: msgpack
+    resource: ingest-events.v1.schema.json
+    examples:
+      - ingest-events/1/
+topic_creation_config:
+  compression.type: lz4
+  retention.ms: "604800000" # 7 days
+  max.message.bytes: "10000000"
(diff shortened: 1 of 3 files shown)


topics/ingest-events-backlog.yaml
```diff
@@ -0,0 +1,17 @@
+Pipeline: errors
+description: Backlog queue for ingest-events
+services:
+  producers:
+    - getsentry/sentry
+  consumers: []
```
lynnagara, line 6: there should be a consuming service here right?

topics/ingest-events-backlog.yaml
```diff
@@ -0,0 +1,17 @@
...
+schemas:
+  - version: 1
+    compatibility_mode: none
+    type: msgpack
+    resource: ingest-events.v1.schema.json
+    examples:
+      - ingest-events/1/
+topic_creation_config:
+  compression.type: lz4
+  retention.ms: "604800000" # 7 days
```
lynnagara, line 16: are we standardising on 7 days for all backlog topics? that may increase the required disk for the cluster a bit. if so, we can probably add a test to this repo that any topic ending in `-backlog` must have 7 day retention

kneeyo1: 7 day retention is taken from DLQ, we haven't really discussed what we should have this be, maybe we should match it with its primary topic?

topics/ingest-events-backlog.yaml
kneeyo1: There wasn't one defined for the transactions backlog topic. But we can define it for both.

topics/ingest-events-backlog.yaml
lynnagara: Primary topic retention seems ok to me - it essentially matches the status quo.

lynnagara (approved):

PR #384, opened 2025-02-14
fix incorrect pipeline annotations

lynnagara:
we found these bugs when using this data to generate initial cluster assignments for itty-bitty


cmanallen (approved):
Approved for replay changes.  But the others look fine as well :D

PR #386, opened 2025-02-14
feat(uptime): Add additional "incident_status" to snuba results

evanpurkhiser:


schemas/snuba-uptime-results.v1.schema.json
```diff
@@ -27,6 +27,10 @@
     "environment": {
       "type": "string",
       "description": "The environment associated with this check"
+    },
+    "in_incident": {
+      "type": "boolean",
+      "description": "Indicates this results is part of a detected incident"
```
wedamija, line 33: Wondering if it makes sense for this to be an int so that we have flexibility if we want to label things differently later?

wedamija (approved):

wedamija (approved):

PR #389, opened 2025-03-05
fix(spans): Align the spans topics with reality

jan-auer:
Various fixes to the `ingest-spans`, `buffered-segments`, and `snuba-spans`
schema definitions:

- Remove `_metrics_summary`, as this is no longer produced and supported
- Make `parent_span_id` and `segment_id` optional everywhere
- Remove `exclusive_time_ms` from `ingest-spans` and `buffered-segments`
- Remove `is_segment` and `segment_id` from `ingest-spans`

Also exposes `SegmentSpan` as named type from `buffered-segments` for the consumer

### DIFF
--- a/schemas/snuba-spans.v1.schema.json
+++ b/schemas/snuba-spans.v1.schema.json
@@ -26,11 +26,11 @@
           "description": "The span ID is a unique identifier for a span within a trace. It is an 8 byte hexadecimal string."
         },
         "parent_span_id": {
-          "type": "string",
+          "type": ["string", "null"],
           "description": "The parent span ID is the ID of the span that caused this span. It is an 8 byte hexadecimal string."
         },
         "segment_id": {
-          "type": "string",
+          "type": ["string", "null"],
           "description": "The segment ID is a unique identifier for a segment within a trace. It is an 8 byte hexadecimal string."
         },
         "profile_id": {
@@ -93,12 +93,6 @@
         },
         "data": {
           "$ref": "#/definitions/Data"
-        },
-        "_metrics_summary": {
-          "type": "object",
-          "additionalProperties": {
-            "$ref": "#/definitions/MetricsSummary"
-          }
         }
       },
       "required": [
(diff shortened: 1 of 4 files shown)


kylemumma (commented):
* What are these files? Are these schema definitions for what data looks like on a kafka topic?
* What is the difference between the 3: buffered_segment, ingest-spans, and snuba-spans? The only topic I am familiar with is the snuba-spans topic that snuba consumes spans off of.

jan-auer:
> * What are these files? Are these schema definitions for what data looks like on a kafka topic?

These files define minimum requirements for valid data on the topics required by the consuming side. The schema can be enforced in consumers. I'm not sure if Snuba uses them, at least Sentry does. @untitaker may have more information on that.

> * What is the difference between the 3: buffered_segment, ingest-spans, and snuba-spans? The only topic I am familiar with is the snuba-spans topic that snuba consumes spans off of.

`snuba-spans` is the only topic we use in production today. Relay produces and Snuba consumes. We're preparing a change in the pipeline that introduces two more intermediate steps, hence the new topics:
- Relay produces to `ingest-spans` and the trace buffer in Sentry consumes.
- The trace buffer produces to `buffered-segments` and a segment processor consumes.
- The segment consumer produces to `snuba-spans`, which Snuba/EAP consume from. 

There are differences between these topics, as each of the pipeline steps adds more attributes to the spans. From Snuba's perspective nothing changes, as Snuba will still consume the same payloads from `snuba-spans`. 

The reason for the change in this schema is that it was incorrect. Specifically, `parent_span_id` and `segment_id` can be `null`. My understanding is that the Snuba consumer already handles this case, as Relay can produce such values, but I'd ask for confirmation on that.

schemas/snuba-spans.v1.schema.json
```diff
@@ -93,12 +93,6 @@
         },
         "data": {
           "$ref": "#/definitions/Data"
-        },
-        "_metrics_summary": {
-          "type": "object",
-          "additionalProperties": {
-            "$ref": "#/definitions/MetricsSummary"
-          }
```
kylemumma, line 101: why is this removed?

schemas/snuba-spans.v1.schema.json
```diff
@@ -209,35 +203,6 @@
...
-        "count": {
-          "type": "number"
-        },
-        "tags": {
-          "type": "object",
-          "additionalProperties": {
-            "type": "string"
-          }
-        }
-      }
```
kylemumma, line 240: why are these removed?

kylemumma (commented):
The change to snuba-spans to allow parent_span_id and segment_id to be null should be safe for snuba consumers because they already handle the null case for these fields. https://github.com/getsentry/snuba/blob/b6f874b16bc533ae233ebfac39110c3d473bc907/rust_snuba/src/processors/spans.rs#L39-L62

schemas/snuba-spans.v1.schema.json
jan-auer: See the PR description - metrics summaries do not exist anymore and will not come back.

Dav1dde (approved):

PR #390, opened 2025-03-12
feat(eap): Add schema for snuba-items

phacops:


untitaker (approved):

volokluev:
Since it's a new topic, wanna define it as protobuf instead of json schema? save us some parsing logic and make things more uniform

k-fish:
Does Relay support sending protobuf to kafka?

schemas/snuba-items.v1.schema.json
```diff
@@ -0,0 +1,60 @@
...
+    "item_id": {
+      "$ref": "#/definitions/UInt"
+    },
+    "item_type": {
+      "type": "integer",
+      "enum": [1, 2, 3]
+    },
+    "attributes": {
+      "$ref": "#/definitions/Attributes"
+    },
```
k-fish, line 24: Do we want to use OTel-ish attribute objects (similar to [ourlogs](https://github.com/getsentry/sentry-kafka-schemas/blob/746ca5aa685f0d64cf7e553f8221196d8cbebd76/schemas/snuba-ourlogs.v1.schema.json#L49-L72) ) or a simple KV? 

I suppose if we do protobuf we don't need to worry about typing w/ json as much

phacops:
> ^ Does Relay support sending protobuf to kafka?

We would have to build a `TraceItem` struct and then send bytes over to Kafka. So, it's "supported" with a bit of elbow grease.

volokluev (approved):

PR #417, opened 2025-06-23
feat(launchpad): Create new kafka schema for launchpad

NicoHinderling:
as suggested here: https://github.com/getsentry/launchpad/pull/16/files#r2162329983


topics/preprod-artifact-upload.yaml
```diff
@@ -0,0 +1,18 @@
+pipeline: generic-metrics
```
NicoHinderling, line 1: I'm not sure what this value should be. Can this just be a new value (ex. `preprod-artifact` or something)?

PR #418, opened 2025-06-23
fix(uptime): Refine new fields on the `uptime-results` schema

wedamija:
Due to how we defined some fields in the schema, the code wasn't generated correctly. Also taking the opportunity to skip allowing fields to nullable. Instead, if the field isn't available, it should just not be set, instead of being null. This aligns with sending data to EAP as well, since EAP doesn't support null attributes.

### DIFF
--- a/schemas/uptime-results.v1.schema.json
+++ b/schemas/uptime-results.v1.schema.json
@@ -58,10 +58,37 @@
       },
       "required": ["start_us", "duration_us"]
     },
-    "NullableTiming": {
-      "oneOf": [{ "$ref": "#/definitions/Timing" }, { "type": "null" }]
+    "RequestDurations": {
+      "title": "request_durations",
+      "description": "Durations of each operation in the request",
+      "type": "object",
+      "properties": {
+        "dns_lookup": {
+          "$ref": "#/definitions/Timing",
+          "description": "Total time from request initiation until DNS resolution completes."
+        },
+        "tcp_connection": {
+          "$ref": "#/definitions/Timing",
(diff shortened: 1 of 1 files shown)


wedamija:
> **versions in use:**
> 
> _The following repositories use one of the schemas you are editing. It is recommended to roll out schema changes in small PRs, meaning that if those used versions lag behind the latest, it is probably best to update those services before rolling out your change._
> 
> **latest version:** 1.3.12
> 
> **changes considered breaking**
> **benign changes**
> ⚠️ **This PR contains breaking changes.** Normally you should avoid that and make your consumer backwards-compatible (meaning that updated consumers can still accept old messages). There are a few exceptions:
> 
> * If consumers already require these invariants in practice, and you're
>   just adjusting the JSON schema to reality, ignore this warning.
> * If you know what you are doing, this change could potentially be rolled out
>   to **producers** first, but that's not a flow we support.

These breaking changes can be ignored, we're not sending these fields

schemas/uptime-results.v1.schema.json
```diff
@@ -58,10 +58,37 @@
...
+          "$ref": "#/definitions/Timing",
+          "description": "Time spent on the TLS handshake."
+        },
+        "time_to_first_byte": {
+          "$ref": "#/definitions/Timing",
+          "description": "Time from request start to receiving the first response header byte."
+        },
+        "send_request": {
+          "$ref": "#/definitions/Timing",
+          "description": "Time spent sending the request (per OTEL conventions)."
```
evanpurkhiser, line 84: Should this also say `per OTEL` and drop the conventions?

PR #420, opened 2025-07-14
feat: Add key_id as required field to ingest-spans and buffered-segments

jan-auer:


Dav1dde (approved):
As discussed marking it required is okay, as long as we make sure Relay is released/deployed first.

As optional, it's always 👍

untitaker (approved):

PR #422, opened 2025-07-16
feat(otlp): Add span links definition to buffered segments

gggritso:
Relay's `SpanKafkaMessage` sends along span links, as the `links` field. The Kafka definition doesn't have links, and this PR adds the definitions.

### DIFF
--- a/schemas/buffered-segments.v1.schema.json
+++ b/schemas/buffered-segments.v1.schema.json
@@ -103,6 +103,13 @@
         },
         "data": {
           "$ref": "#/definitions/Data"
+        },
+        "links": {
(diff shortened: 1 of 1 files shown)


untitaker:
if you expand the `changes considered breaking` you will see:

> Added a new property, but the consumer has been ignoring additional properties so far. This is probably still fine, but please double-check that the producer does not already send this property with a different type in practice than you defined in this schema.

i think your reasoning makes sense fwiw, these messages are just there to make ppl think about backwards compat at all

Dav1dde (approved):
This matches the current format Relay already emits

PR #424, opened 2025-07-16
feat(tasks): Add taskworker-workflows-events kafka topic

kcons:
Add a new Kafka topic to be used by process_workflows_event and other similar high-volume workflow processing events.


CODEOWNERS
```diff
@@ -119,6 +119,7 @@
 /topics/taskworker-email.yaml                                 @getsentry/taskbroker
 /topics/taskworker-email-dlq.yaml                             @getsentry/taskbroker
 /topics/taskworker-cutover.yaml                               @getsentry/taskbroker
+/topics/taskworker-workflows-events.yaml                      @getsentry/taskbroker
```
markstory, line 122: What about `workflow-engine` so it is aligned with the task namespace that will run here?

kcons: I'm not sure I understand the question, and to the extent the question is about the purpose of the topic, should I be using the topic description (i'd noticed many descriptions are identical so I followed suit).
The comment here being on codeowners is also making me wonder if I'm understanding right.

My theory for this topic is simple: we intend to do workflow-engine-related post processing on nearly all error events. The scale and uniformity of that workload is different from our delayed processing, action execution, and probably activity-based tasks, so this topic is for isolating that work. "It's for our events tasks", in other words.

Or should I be reusing errors ingest topics here, as that's the domain we're describing.

I can see another argument for us to have a shared pool for all workflow-related execution; from a "team that owns X monitors/manages X worker capacity" angle, that has an appeal.

I'm not sure what the idiom is here; this is my first time interacting with our topics, and I'm mostly working from the 'Monitoring Tasks' doc.

evanh (approved):

PR #427, opened 2025-07-22
feat: snuba-items items dlq topic definition

kylemumma:
this is the topic definition for a new dlq on snuba-items consumer


topics/snuba-dead-letter-items.yaml
```diff
@@ -0,0 +1,16 @@
+pipeline: items
```
kylemumma, line 1: snuba-items.yaml doesnt have this pipeline but snuba-items-commit-log.yaml does. not sure if i should have it

phacops (approved):

PR #434, opened 2025-08-27
Add replay_relay_snuba_publish_disabled field to recording schema

cmanallen:
Related: https://github.com/getsentry/relay/pull/5088
Related: https://github.com/getsentry/sentry/pull/98353


cmanallen:
@untitaker Ah understood. Did not see the toggle there. I'm certain its not being used.

untitaker:
if it's temporary you don't need to add it to begin with

i don't know if `default: true` is great because `Default for bool` is `false` in rust

PR #435, opened 2025-08-27
ref(schemas-diff): Expand details by default

untitaker:
I've been asked twice by different ppl why a schema is breaking, and it
seems that the message is too dense or poorly formatted to drive the
point home. Expand details and hope that helps


cmanallen (approved):
Is the message too dense or the people 🤔

untitaker:
> Is the message too dense or the people 🤔

not sure, looking it at it now i feel that the formatting makes it somehow difficult to read.

PR #437, opened 2025-09-17
feat(preprod): add requested_features field to PreprodArtifact event schema

trevor-e:
Adds an optional `requested_features` string array field to our schema so we can control which analysis features to run.

edit: see my comment below about the flagged breaking change


trevor-e:
I believe ^ should be safe. We are already producing this value (https://github.com/getsentry/sentry/blob/master/src/sentry/preprod/producer.py#L53) but the type signature should match.

NicoHinderling (approved):

PR #480, opened 2026-04-08
feat(tasks) Add topic for seer taskworker

markstory:
Seer is will be moving to taskbroker for its internal tasks and tasks spawned from sentry.

Refs STREAM-864


evanh (approved):
One small step for Sentry, one giant leap towards no more Celery!

PR #487, opened 2026-06-12
feat(profiles): Add `attachments` to profiles schema

markushi:
Adds an optional `attachments` array to the profiles message. Each entry describes a file related to the profile chunk (e.g. a raw perfetto profile) that is stored in the object store, with a `name`, an optional `content_type`, and the `stored_id` referencing the stored object.

The field is additive and optional, so the change is backward compatible for existing producers and consumers. A new msgpack example (`examples/profiles/1/example_with_attachments.msgpack`) exercises the field; it is picked up automatically since the topic references the whole `profiles/1/` examples directory.


schemas/profiles.v1.schema.json
```diff
@@ -13,6 +13,19 @@
...
       "description": "bytes"
+    },
+    "attachments": {
+      "description": "files related to the profile chunk (e.g. a raw profile), stored in the object store",
+      "type": "array",
+      "items": {
+        "type": "object",
+        "required": ["name", "stored_id"],
+        "properties": {
+          "name": { "type": "string" },
```
Dav1dde, line 24: What would we send the `name` in Relay?

markushi: I think Joris mainly added this to be on-par with event attachments, where name implies the file name. In our case this could be a static const string defined in relay, e.g. `raw_profile.pftrace`

Dav1dde: Then maybe we don't make it optional.

markushi: 👍 yeah, that's why I marked it as required already.

Dav1dde (approved):

PR #489, opened 2026-06-18
fix(transactions): allow nullable measurements, breakdowns, and measurement entries

phacops:
## Summary

Relay replaces invalid SDK-supplied field values with `null` rather than omitting them. The transactions schema did not account for this, causing validation failures when `measurements` or `breakdowns` arrive as `null`.

- `measurements` field in `Data` is now nullable (`anyOf: [Measurements, null]`)
- `breakdowns` field in `Data` is now nullable (`anyOf: [Breakdowns, null]`)
- Individual entries inside `Measurements` are now nullable (`additionalProperties` allows `NumOfSpans | null`, and `num_of_spans` itself is `NumOfSpans | null`)

Addresses review comment: https://linear.app/getsentry/review/fixtransactions-default-null-measurements-to-an-empty-object-a72e94787c66#comment-73e9ca52

## Test plan

- [x] All existing example validation tests pass (`549 passed`)

https://claude.ai/code/session_01Cffb1Z7voprAuHqnwNhYGv

---
_Generated by [Claude Code](https://claude.ai/code/session_01Cffb1Z7voprAuHqnwNhYGv)_

### DIFF
--- a/schemas/transactions.v1.schema.json
+++ b/schemas/transactions.v1.schema.json
@@ -298,10 +298,16 @@
           }
         },
         "measurements": {
-          "$ref": "#/definitions/Measurements"
+          "anyOf": [
+            { "$ref": "#/definitions/Measurements" },
+            { "type": "null" }
+          ]
         },
         "breakdowns": {
-          "$ref": "#/definitions/Breakdowns"
+          "anyOf": [
+            { "$ref": "#/definitions/Breakdowns" },
+            { "type": "null" }
+          ]
         },
         "culprit": {
           "type": "string"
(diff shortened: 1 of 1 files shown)


Dav1dde (approved):

schemas/transactions.v1.schema.json
```diff
@@ -859,9 +862,12 @@
     },
     "Measurements": {
       "type": "object",
```
Dav1dde: In spans this is typed as:

```
    "Measurements": {
      "type": ["object", "null"],
      "properties": {
        "num_of_spans": {
          "$ref": "#/definitions/NumOfSpans"
        }
      }
```

No clue if that makes an actual difference.

phacops: Good call — switched to that convention. `Measurements` and `Breakdowns` are now typed as `["object", "null"]` on the definitions themselves, and the field references are plain `$ref`s again (no `anyOf` wrapper). This matches the spans schema and also makes the `json-schema-changes` diff cleaner (it now resolves the `$ref` to a simple `object → [object, null]` type addition rather than reporting a bunch of spurious type additions). I also dropped the per-entry nullability I'd added on `num_of_spans`/`additionalProperties`, since the spans pattern keeps those as plain `NumOfSpans` and the actual need is just the top-level object being null. All 676 tests pass.

---
_Generated by [Claude Code](https://claude.ai/code)_

PR #491, opened 2026-06-23
feat(loadbalancer-outcomes): Add loadbalancer-outcomes topic

Dav1dde:
E.g. https://github.com/getsentry/ops/blob/master/shared_config/kafka/topics/regional_overrides/us/loadbalancer-outcomes.yaml

Checked the generated output of topicctl for the `topic_creation_config`.


fpacifici (changes requested):
You do not need to add a topic here if it is not meant to run self hosted. 
You can simply add it in the ops repo definition. like we did for ingest-events-2 
https://github.com/getsentry/ops/blob/master/shared_config/kafka/topics/regional_overrides/us/ingest-events-2.yaml

Dav1dde:
Discussed in Slack. There is an alternative solution for the data team.

PR #495, opened 2026-07-14
feat(taskbroker): Remove Private Taskworker Topics

george-sentry:
## Linear
Refs [STREAM-1238](https://linear.app/getsentry/issue/STREAM-1238/have-taskbroker-use-private-kafka-topics)

## Description
Now that `ops` uses either `taskworker` or `taskworker-dlq` to configure every taskworker topic, we can remove all the other taskworker topics from this repository are they are now "private."


topics/taskworker-control.yaml
```diff
@@ -1,20 +0,0 @@
-pipeline: taskworker
```
bmckerry, line 1: not sure if we want to keep this one for now

bmckerry (approved):

enochtangg (approved):

topics/taskworker-control.yaml
george-sentry: Added it back.

PR #502, opened 2026-09-11
feat(ingest-monitors): add optional partition_ids to the clock pulse message

strongs:
First step in making monitor clock state safe to lose. [Context](https://sentry.slack.com/archives/C0BCRGVJ8F5/p1789081921128639).

The monitors clock needs to know which partitions to expect for each tick. The plan is for the minute pulse to carry that list. This keeps the Kafka metadata read in the scheduled `clock_pulse` task and off the check-in path.

btw if you want to decode the binary yourself, run
```
python -c "import msgpack; print(msgpack.unpackb(open('examples/ingest-monitors/1/clock_pulse_with_partitions.msgpack','rb').read()))"
```

Notes on rollout:
1. Merge this PR first
2. Release a new kafka-schemas version
3. In `sentry`, change to the new version. Do not change the producer yet
4. Ensure every consumer is running the new version
5. Change producers so that the new field `partition_ids` is populated

Refs INFRENG-625


strongs:
Failing check appears to be an issue with json-schema-diff, it doesn't support `oneOf`. Doesn't seem problematic.

ldelvoye (approved):

wedamija (approved):

markstory (approved):

